PR #14, opened 2024-04-16
[16.0][MIG] stock_orderpoint_mto_as_mts

tuantrantg:
Depends on:
* https://github.com/OCA/stock-logistics-orderpoint/pull/5

Notes:
* Renamed sale_stock_mto_as_mts_orderpoint to stock_orderpoint_mto_as_mts
* Removed dependency on stock_orderpoint_manual_procurement as not needed anymore, see https://github.com/OCA/stock-logistics-warehouse/pull/1641
* Used `is_mto` on product (depend on product-attribute/product_route_mto)
* Used `is_mto` on route instead of xmlid (provided by stock-logistics-warehouse/stock_route_mto - already pulled by product_route_mto)
* Performed the bellow in `product.product`:
  * Processed: the orderpoint must be created only for warehouses where there is no mto rule for that warehouse
  * Moved the orderpoint archiving from `product.template` and refactored it
* Made the orderpoint archiving configurable on the warehouse
* Dropped `is_from_mto_route` on `stock.move`. That should not be in this module, not related to orderpoint


stock_orderpoint_mto_as_mts/models/stock_warehouse.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class StockWarehouse(models.Model):
+
+    _inherit = "stock.warehouse"
+
+    archive_orderpoints_mto_removal = fields.Boolean(default=False)
+
+    def _get_locations_for_mto_orderpoints(self):
+        return self.mapped("lot_stock_id")
```
jbaudoux: Can you depend on https://github.com/OCA/stock-logistics-orderpoint/pull/5/files#diff-34ce65f805726da9a9ebaedb97f0d4188d1335149786f5d37cf07008811ea0ddR11 ?

```suggestion
        self.ensure_one()
        return self.default_orderpoint_location_id or self.lot_stock_id
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,141 @@
...
+            product.is_missing_default_orderpoint_for_mto = (
+                product.is_mto
+                and product.type == "product"
+                and len_wh != len_wh_orderpoint
+            )
+
+    def _create_default_orderpoint_for_mto(self):
+        default_company = self.env["res.company"]._get_main_company()
+        for company, products in self.partition("company_id").items():
+            company = company or default_company
```
jbaudoux: If there is no company, then the product is valid for all companies, not just the main one. You then need to get all warehouses

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,141 @@
...
+            warehouses = self.env["stock.warehouse"].search(
+                [("company_id", "=", company.id)]
+            )
+            for product in products:
+                if not product.is_missing_default_orderpoint_for_mto:
```
jbaudoux: This needs to be refactored as you need to know for which warehouse it is missing. Maybe drop the computed field

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,141 @@
...
+                product.is_mto
+                and product.type == "product"
+                and len_wh != len_wh_orderpoint
+            )
+
+    def _create_default_orderpoint_for_mto(self):
+        default_company = self.env["res.company"]._get_main_company()
+        for company, products in self.partition("company_id").items():
+            company = company or default_company
+            warehouses = self.env["stock.warehouse"].search(
```
jbaudoux: You need to filter out the warehouses for which we want mto as mts. Ideally as an option on the warehouse.

jbaudoux (changes requested):

stock_orderpoint_mto_as_mts/models/product_product.py
tuantrantg: Hi @jbaudoux please review these again

santostelmo (approved):
LG good with @jbaudoux changes

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,123 @@
...
+    def _get_warehouse_missing_orderpoint_for_mto(self):
+        res = {}
+        wh_obj = self.env["stock.warehouse"]
+        for product in self:
+            wh_domain = [("mto_as_mts", "=", True)]
+            if product.company_id:
+                wh_domain.append(("company_id", "=", product.company_id.id))
+            wh = wh_obj.search(wh_domain)
+            wh_orderpoint = product.orderpoint_ids.partition("warehouse_id").keys()
+            wh_wo_orderpoint = wh.filtered(lambda r: r not in wh_orderpoint)
```
jbaudoux: You mean?
```suggestion
            wh_orderpoint = product.orderpoint_ids.warehouse_id
            wh_wo_orderpoint = wh - wh_orderpoint
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,123 @@
...
+            orderpoint = orderpoint_obj.with_context(active_test=False).search(
+                [
+                    ("product_id", "=", self.id),
+                    (
+                        "location_id",
+                        "=",
+                        warehouse._get_locations_for_mto_orderpoints().id,
+                    ),
+                ],
+                limit=1,
```
jbaudoux, line 41: To prevent useless sorting cost, always better to set order when limit=1
```suggestion
                limit=1, order="id"
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,123 @@
...
+from odoo import api, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_warehouse_missing_orderpoint_for_mto(self):
+        res = {}
+        wh_obj = self.env["stock.warehouse"]
+        for product in self:
```
jbaudoux, line 14: Maybe use ensure_one() and return wh_wo_orderpoint instead of dict

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-stock-orderpoint-default-location @ git+https://github.com/OCA/stock-logistics-orderpoint.git@refs/pull/5/head#subdirectory=setup/stock_orderpoint_default_location
```
jbaudoux: This can now be dropped

stock_orderpoint_mto_as_mts/models/product_product.py
santostelmo, line 14:
```suggestion
        for product in self:
            if not product.purchase_ok:
                continue
```
There's no point to create a orderpoints for products that can not be purchased.

That is in conflict with the validation done in `purchase_reorder_control`
_Unpurchaseable Products cannot be reordered_
https://github.com/OCA/purchase-workflow/blob/e113f0a3a046ad832f2e324df50301c276932fa4/purchase_reorder_control/models/stock_warehouse_orderpoint.py#L21

santostelmo (changes requested):

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_warehouse_missing_orderpoint_for_mto(self):
+        self.ensure_one()
+        res = False
+        wh_obj = self.env["stock.warehouse"]
+        for product in self:
```
jbaudoux, line 14: drop for loop as you have `ensure_one()`

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,128 @@
...
+            "product_uom": self.uom_id.id,
+        }
+
+    def _ensure_default_orderpoint_for_mto(self):
+        """Ensure that a default orderpoint is created for the MTO products.
+
+        that have no orderpoint yet.
+        """
+        for product in self:
+            wh = product._get_warehouse_missing_orderpoint_for_mto()
```
jbaudoux, line 72: I don't see where you check it is a mto product
```
mto_products = self._filter_mto_products()
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,128 @@
...
+                product._create_default_orderpoint_for_mto(wh)
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        products = super().create(vals_list)
+        products.sudo()._ensure_default_orderpoint_for_mto()
+        return products
+
+    def write(self, vals):
+        # Archive orderpoints when MTO route is removed
```
jbaudoux, line 83: this comment is misplaced

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,128 @@
...
+            if wh:
+                product._create_default_orderpoint_for_mto(wh)
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        products = super().create(vals_list)
+        products.sudo()._ensure_default_orderpoint_for_mto()
+        return products
+
+    def write(self, vals):
```
jbaudoux, line 82: You need to look in`vals` for `is_mto` or `company_id`. Those are on `product.template`, so you need to move the hook to write on the right model

stock_orderpoint_mto_as_mts/models/stock_warehouse.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+
+
+class StockWarehouse(models.Model):
+
+    _inherit = "stock.warehouse"
+
+    mto_as_mts = fields.Boolean(default=False)
```
jbaudoux, line 10: When you flag this, you could archive from the MTO route the rule for this warehouse

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: I don't really understand the method
```
res = super().write(vals)
if vals.get("is_mto"):
    # when is_mto is set
    self.sudo()._ensure_default_orderpoint_for_mto()
elif not vals.get("is_mto", True):
    # when is_mto is removed
    self.sudo()._archive_orderpoints_on_mto_removal()
elif "company_id" in vals:
    # when company is changed
    self.sudo()._archive_orderpoints_on_mto_removal()
    self.sudo()._ensure_default_orderpoint_for_mto()
return res
```

stock_orderpoint_mto_as_mts/models/stock_warehouse.py
jbaudoux: `default=False` is useless

PR #4, opened 2023-07-13
[ADD] stock_location_orderpoint_average_daily_sale

rousseldenis:
Depends on:

- #3 
- https://github.com/OCA/stock-logistics-reporting/pull/257

PR #15, opened 2024-04-27
[16.0][IMP] stock_location_orderpoint: Add min_max replenish_rule

mt-software-de:
cc @rousseldenis @jbaudoux

### DIFF
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -57,11 +57,15 @@ class StockLocationOrderpoint(models.Model):
         default="fill_up",
         required=True,
         help="Defines how the qty to replenish gets computed\n"
-        "Fill up = The replenishment will be triggered when a move is waiting availability "
-        "and forecast quantity is negative at the location (i.e. min=0). "
-        "The replenished quantity will bring back the forecast quantity to 0 (i.e. max=0) "
+        "Fill up = Forecast quantity must be less then the min_qty at the location "
+        "The replenished quantity will bring back the forecast quantity to max_qty "
         "but will be limited to what is available at the source location "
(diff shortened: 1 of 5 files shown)


stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -53,7 +53,7 @@ class StockLocationOrderpoint(models.Model):
         "Scheduled orderpoints are triggered via scheduled actions per location",
     )
     replenish_method = fields.Selection(
-        [("fill_up", "Fill up")],
+        [("fill_up", "Fill up"), ("min_max", "Min Max")],
```
rousseldenis: Ugggh. Really?

IMHO, this should go in another module and usage/description should be completed in order to explain the use case as it sounds strange for me.

mt-software-de: After thinking about it would suggest to extend the fill_up, make it work with min_qty and max_qty by default 0.
Also i would than refactor the part, where the consuming move domain is generated. https://github.com/OCA/stock-logistics-orderpoint/blob/0209931876c9d811d06b3402082a0c3d395e148b/stock_location_orderpoint/models/stock_location_orderpoint.py#L199 to make in configurable. 

This would then fit the needs of my customer. 
The use i am facing is.
A sale is happening and the qty in on specific location is less than the want.
As example the want to have always 10 pieces of productA on the location Shop. 
After a sale of productA the available qty decreased on the Shop location to a qty of 4. Then the want to bring 6 pieces of productA to the shop location.

mt-software-de: Is just a simpler uses case of your PR https://github.com/OCA/stock-logistics-orderpoint/pull/4/.

Also this can be maybe later used for replenish via purchase.

PR #19, opened 2024-05-27
[16.0][MIG] stock_orderpoint_mto_as_mts: Migration to 16.0

QuocDuong1306:
This PR takes over:
- https://github.com/OCA/stock-logistics-orderpoint/pull/14

Notes:

- Renamed `sale_stock_mto_as_mts_orderpoint` to `stock_orderpoint_mto_as_mts`
- Removed dependency on stock_orderpoint_manual_procurement as not needed anymore, see https://github.com/OCA/stock-logistics-warehouse/pull/1641
- Used `is_mto` on product (depend on product-attribute/product_route_mto)
- Used `is_mto` on route instead of xmlid (provided by stock-logistics-warehouse/stock_route_mto - already pulled by product_route_mto)
- Performed the bellow in `product.product`:
- Processed: the orderpoint must be created only for warehouses where there is no mto rule for that warehouse
- Moved the orderpoint archiving from product.template and refactored it
- Made the orderpoint archiving configurable on the warehouse
- Dropped `is_from_mto_route` on stock.move. That should not be in this module, not related to orderpoint


QuocDuong1306:
Hi @jbaudoux, I applied your suggestions in https://github.com/OCA/stock-logistics-orderpoint/pull/14 to the last commit
BTW, added some tests also

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+            return
+        elif "route_ids" not in vals:
+            # No changes about is_mto
+            self.sudo()._ensure_default_orderpoint_for_mto()
+            return
+        mto_products = self._filter_mto_products()
+        # Find all MTO products that have been changed to Not MTO
+        products_to_update = original_mto_products - mto_products
+        if products_to_update:
+            products_to_update._archive_orderpoints_on_mto_removal()
```
jbaudoux: This is only necessary when is_mto is removed
```suggestion
        if not vals.get("is_mto", True):
            # when is_mto is removed
            mto_products = self._filter_mto_products()
            # Find all MTO products that have been changed to Not MTO
            products_to_update = original_mto_products - mto_products
            if products_to_update:
                products_to_update._archive_orderpoints_on_mto_removal()
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+        return
+
+    def _filter_mto_products(self):
+        return self.filtered(lambda p: p.is_mto)
```
jbaudoux: This method is useless. There is already the `is_mto` compute method

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+        return res
+
+    def _update_mto_products(self, original_mto_products, vals):
+        if "company_id" in vals:
+            # Change company, must archive orderpoints
+            self.sudo()._archive_orderpoints_on_mto_removal()
+            self.sudo()._ensure_default_orderpoint_for_mto()
+            return
+        elif "route_ids" not in vals:
+            # No changes about is_mto
```
jbaudoux:
```suggestion
        if vals.get("is_mto"):
            # when is_mto is set
```

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+        products.sudo()._ensure_default_orderpoint_for_mto()
+        return products
+
+    def write(self, vals):
+        original_mto_products = self._filter_mto_products()
+        res = super().write(vals)
+        self._update_mto_products(original_mto_products, vals)
+        return res
+
+    def _update_mto_products(self, original_mto_products, vals):
```
jbaudoux: I would move this to product.template

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+    def create(self, vals_list):
+        products = super().create(vals_list)
+        products.sudo()._ensure_default_orderpoint_for_mto()
+        return products
+
+    def write(self, vals):
+        original_mto_products = self._filter_mto_products()
+        res = super().write(vals)
+        self._update_mto_products(original_mto_products, vals)
+        return res
```
jbaudoux: Can be dropped, `is_mto` field is on template

stock_orderpoint_mto_as_mts/models/product_template.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        template = super().create(vals_list)
+        template.product_variant_ids.sudo()._ensure_default_orderpoint_for_mto()
+        return template
```
jbaudoux, line 14: Can be dropped. Orderpoint is on product

jbaudoux (changes requested):
As we rely on module `product_route_mto` we should not re-implement the concept of `is_mto`. This is currently defined on product.template. In my opinion, there is extraneous complexity to cover the case where only a variant would be mto. It cannot be so far.

stock_orderpoint_mto_as_mts/models/product_product.py
QuocDuong1306: Hello @jbaudoux , as I knew and tested, `is_mto` is the computed field, so it didn't appear in `vals` of `def write()`
So `not vals.get("is_mto", True)` is not worked in this case, that is the reason why I used `route_ids` instead of `is_mto`

jbaudoux: It is defined on product.template and it appears on write vals of product.template but not on product.product as it is not defined there. That's why it is useless to hook on product.product.

QuocDuong1306: Yes, I know it, [It is defined on product.template](https://github.com/OCA/product-attribute/blob/3a646a09de5e0bc90f3a59328396b2a3da9d4771/product_route_mto/models/product_template.py#L11), but it is the computed field (and readonly = True in this case), if we want to change it, we must write on the dependency field (`route_ids`, `categ_id` and `is_mto of the route` in this case)

After that, `is_mto` on product.template will appear in vals of `def _write()`, not in `def write`

stock_orderpoint_mto_as_mts/models/product_template.py
QuocDuong1306: If we drop this, will fail on case: Create product.template --> Create orderpoints

Because when create template --> template will trigger [here](https://github.com/odoo/odoo/blob/e3033d7aae5537403a1aedbae6e598514c04f703/addons/product/models/product_template.py#L433) --> for now, products will be created but `is_mto` is not computed (debugged) so we can not handle this case through `def create() of product.product`

QuocDuong1306:
Hi @jbaudoux , I updated the PR, and added more logics to handle cases that impact to `is_mto` on product (`stock.route`, `product.category`)

Based on [here](https://github.com/OCA/product-attribute/blob/3a646a09de5e0bc90f3a59328396b2a3da9d4771/product_route_mto/models/product_template.py#L16)

cyrilmanuel (approved):

jbaudoux (approved):

jbaudoux:
cc @rousseldenis

cyrilmanuel:
Hi @rousseldenis  is it good for you ?

stock_orderpoint_mto_as_mts/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,13 @@
+This module aims to materialize the triggering of another stock rule through
+an orderpoint in the procurement of MTO products from a sale order.
```
jbaudoux:
```suggestion
an orderpoint in the procurement of MTO products.
```

stock_orderpoint_mto_as_mts/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,13 @@
...
+When a sales order is confirmed, if the delivery order move is generated by
+the stock rule linked to the standard MTO route, and the MTO route is marked
+on the product, an orderpoint will be created on the Stock location of the
+warehouse, with min/max quantities of zero.
+
+This allows to regenerate procurement according to the procure_recommended_qty
+using module `stock_orderpoint_manual_procurement`.
+
+Finally, orderpoints with min/max quantities of zero will be archived if the
+MTO route is removed on the product.
```
jbaudoux:
```suggestion
The MTO route will loose it's MTO rule as this is replaced by the generated orderpoints.

When a product is configured as MTO, then an orderpoint with min/max of zero is created. When the product is not anymore configured as MTO, the orderpoint can be automatically archived depending on the configuration you choose on the warehouse.
```

stock_orderpoint_mto_as_mts/readme/CONFIGURATION.rst
```diff
@@ -0,0 +1,12 @@
...
+
+* Keep the triggering of another rule as Supply Method on stock.rule, will
+  ensure that the need is materialized by the procure_recommended_qty on the
+  orderpoint, if the ensuing procurement (purchase order/manufacturing order)
+  having the MTO rule as origin is canceled.
+
+* Change the Supply Method on stock.rule to Take From Stock, will ensure that
+  the need is materialized by the procure_recommended_qty on the orderpoint,
+  if the ensuing procurement (purchase order/manufacturing order) having the
+  orderpoint as origin is canceled.
```
jbaudoux:
```suggestion
In order to modify the MTO route and rely on generated orderpoints for MTO products instead of the MTO rule, you need to enable "mto as mts" on the warehouse
```

jbaudoux (commented):
Can you update the documentation with the performed modifications?

stock_orderpoint_mto_as_mts/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Sale Stock Mto As Mts Orderpoint",
```
TDu:
```suggestion
    "name": "Stock Orderpoint Mto As Mts",
```

QuocDuong1306:
Hi guys, the PR is updated

PR #22, opened 2024-08-13
[16.0][MIG]stock_orderpoint_origin_mrp_link: migration to `16.0`

PicchiSeba:
This PR ports `stock_orderpoint_origin_mrp_link` to `16.0` (It also ports `stock_orderpoint_origin` because a single commit added both modules)

These modules come from https://github.com/OCA/stock-logistics-warehouse/tree/14.0/


HekkiMelody (approved):
Code review, LGTM

giarve:
Interesting... if compatible with
https://github.com/OCA/purchase-workflow/pull/1983
and its equivalent for mrp, it would make Odoo with orderpoint MTS replenishment equivalent to MTO with more flexibility (without the need for default MTS + MTO route)

PicchiSeba:
Hi @giarve , I'm not familiar with the module you linked. What do you propose to increase compatibility?

giarve:
> Hi @giarve , I'm not familiar with the module you linked. What do you propose to increase compatibility?

I have not tried them both at the same time yet. Theoretically they should both work because origin is split by **,** and that's what the module uses to find the source (origin) orders. 

What I thought is interesting is that when using this module and filling the origin field, an equivalent module to "purchase_sale_link_by_origin" for mrp, could directly link manufacturing orders generated from the orderpoint to the sales order even when not using MTO route (and using the replenishment window).

PicchiSeba:
Hello @giarve , I finally had the time to take a look at https://github.com/OCA/purchase-workflow/pull/1983. But it seems they do things differently.

I would keep them separated, at least for this migration.
We might want to change the behaviour in the future once this module is in a stable state.

PicchiSeba:
PR updated with a couple of missed commits

giarve (approved):

HekkiMelody:
@OCA/logistics-maintainers We've been using these ports in prod for months now.

hparfr (approved):
/ocabot merge nobump

PR #28, opened 2025-01-13
[18.0][MIG] sale_stock_mto_as_mts_orderpoint: Migration to 18.0

chaule97:
- move from `OCA/stock-logistics-workflow` to `OCA/stock-logistics-orderpoint` 
-  because `stock_orderpoint_manual_procurement` is [deprecated](https://github.com/OCA/stock-logistics-warehouse/pull/1641#issuecomment-1445054077), so I remove it in depends and write more code about `stock_warehouse_orderpoint`


sale_stock_mto_as_mts_orderpoint/models/stock_warehouse.py
```diff
@@ -9,21 +9,25 @@ class StockWarehouse(models.Model):
...
+        domain = [
+            ("route_id", "=", mto_route.id),
+            "|",
+            ("warehouse_id", "in", warehouses_with_mto.ids),
+            ("picking_type_id.warehouse_id", "in", warehouses_with_mto.ids),
+        ]
+
+        wh_mto_rules = self.env["stock.rule"].search(domain)
+        if wh_mto_rules:
+            wh_mto_rules.write({"active": False})
```
grindtildeath, line 33: Please avoid rewriting for cosmetic changes in migration

sale_stock_mto_as_mts_orderpoint/models/sale_order.py
```diff
@@ -0,0 +1,99 @@
...
+        return res
+
+    def _run_orderpoints_for_mto_products(self):
+        mto_route = self.env.ref("stock.route_warehouse0_mto", raise_if_not_found=False)
+        if not mto_route:
+            return
+
+        orderpoints_to_procure_ids = []
+        for line in self:
+            delivery_moves = line.move_ids.filtered(
```
grindtildeath, line 23: :warning: This won't work with new push rules that do not necessarily create the delivery move

grindtildeath (changes requested):
Based on the new execution of procurement and delivery moves not being created on the confirmation of the order, this will not work.

I'll get back to it later as I'm not sure yet how we should do this.

jbaudoux (changes requested):
You need to restart and migrate https://github.com/OCA/stock-logistics-orderpoint/tree/16.0/stock_orderpoint_mto_as_mts

jbaudoux:
History of changes is here https://github.com/OCA/stock-logistics-orderpoint/pull/19

PR #33, opened 2025-02-06
[MIG] stock_orderpoint_route: Migration to 18.0

grindtildeath:


ivantodorovich (approved):

grindtildeath:
This one will need fixing because IMO the condition to filter routes in `_get_location_routes_of_parents` is not correct for following reasons:

- In v18.0, multi step reception route is excluded because its rule doesn't use an internal source location
- Even in v16.0, using a single step reception route, it is excluded because its single rule doesn't use an internal source location
- The delivery route should not be selectable from the orderpoint as triggering the procurement with this route selected will not generate anything

giarve:
> This one will need fixing because IMO the condition to filter routes in `_get_location_routes_of_parents` is not correct for following reasons:
> 
>     * In v18.0, multi step reception route is excluded because its rule doesn't use an internal source location
> 
>     * Even in v16.0, using a single step reception route, it is excluded because its single rule doesn't use an internal source location
> 
>     * The delivery route should not be selectable from the orderpoint as triggering the procurement with this route selected will not generate anything

Check this out:

https://github.com/OCA/stock-logistics-orderpoint/pull/42

marcelsavegnago (approved):
LGTM

marcelsavegnago:
@grindtildeath can you rebase?

PR #37, opened 2025-05-27
[18.0][MIG] stock_orderpoint_no_horizon: Migration to 18.0

henrybackman:


henrybackman:
The tests still don't pass on CI, but they work for me locally so not certain how to proceed with it

jbaudoux:
@henrybackman install stock_orderpoint_purchase_link and test will fail. Can you investigate?

henrybackman:
I looked into the tests and they fail when `puchase_stock` module is installed but it wasn't clear to me how this module should work with `purchase_stock` as it changes how the stock moves are generated

jbaudoux:
@henrybackman I pushed a fix. I also removed the delay on the rule, otherwise it was inside the horizon

simahawk (approved):

PR #39, opened 2025-06-07
[18.0][MIG] stock_orderpoint_mto_as_mts: Migration to v18.0 + IMP

jbaudoux:
Depends on:
* https://github.com/OCA/stock-logistics-orderpoint/pull/40

Tested

Improvements:
* Improve performance on each product create/write
* Fix changing a product category as the route is propagated to children categories
* Allow to process lots of products linked to a product category change
* Better manage multi-company


stock_orderpoint_mto_as_mts/models/stock_route.py
```diff
@@ -0,0 +1,32 @@
...
+
+class StockRoute(models.Model):
+    _inherit = "stock.route"
+
+    def write(self, vals):
+        if "is_mto" not in vals:
+            return super().write(vals)
+        products = self.product_ids.product_variant_ids
+        products |= self.env["product.product"].search(
+            [("categ_id", "child_of", self.ids)]
```
simahawk: searching for products using route ids instead of categories' seems wrong :thinking:

simahawk (changes requested):

stock_orderpoint_mto_as_mts/tests/test_stock_warehouse_mto_as_mts.py
```diff
@@ -1,10 +1,11 @@
 # Copyright 2024 Camptocamp SA
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
-from odoo.tests import common
+from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestStockWareHouseMtoAsMts(common.TransactionCase):
```
ivantodorovich: You can drop this from a few lines below

```diff
-cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

stock_orderpoint_mto_as_mts/models/product_category.py
```diff
@@ -1,34 +1,45 @@
...
-
-        if not vals.get("route_ids", False):
-            # No changes for MTO routes, so no changes for is_mto on products
-            res = super().write(vals)
-            self.product_ids.sudo()._ensure_default_orderpoint_for_mto()
-            return res
-
-        original_mto_products = self.product_ids.filtered(
-            lambda product: product.is_mto
+        if not self or "route_ids" not in vals:
```
ivantodorovich: Looking at what the code does, `parent_id` should also be a condition here, as it can have an effect on `total_route_ids`

stock_orderpoint_mto_as_mts/models/product_category.py
```diff
@@ -1,34 +1,45 @@
...
+        # Update orderpoints for MTO / not MTO products that are linked to
+        # these categories. A category is not bound to a company, search
+        # products accross all companies.
+        _logger.debug("Search products already as MTO")
+        products = (
+            self.env["product.product"]
+            .sudo()
+            .search([("categ_id", "child_of", self.ids)])
         )
+        original_mto_products = products.filtered_domain([("is_mto", "=", True)])
```
ivantodorovich: FTR, using `filtered_domain` (now) vs `filtered` (before) is actually worse perf-wise.

In any case, `is_mto` is stored.. why not adding it to the `search` domain directly?

stock_orderpoint_mto_as_mts/models/product_category.py
```diff
@@ -1,34 +1,45 @@
...
-
-        original_mto_products = self.product_ids.filtered(
-            lambda product: product.is_mto
+        if not self or "route_ids" not in vals:
+            return super().write(vals)
+
+        # Update orderpoints for MTO / not MTO products that are linked to
+        # these categories. A category is not bound to a company, search
+        # products accross all companies.
+        _logger.debug("Search products already as MTO")
```
ivantodorovich: What's the deal with all these debug comments?
Left-over from debugging session? or do we really want to keep them?

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -1,58 +1,44 @@
...
-                ],
-                limit=1,
-                order="id",
+        warehouses = self.env["stock.warehouse"].search(
+            [("mto_as_mts", "=", True), ("archive_orderpoints_mto_removal", "=", True)]
+        )
+        if warehouses:
+            # Find all MTO products that changed to non MTO
+            _logger.debug("Search products not MTO anymore")
+            products_not_mto = original_mto_products.filtered_domain(
```
ivantodorovich: Same comment as before:
- use `filtered_domain` only when you truly need it, otherwise prefer `filtered` for perf
- if you can avoid it, especially when dealing with a potentially large number of records, filter them with search instead

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -1,58 +1,44 @@
...
-        wh_domain = [("mto_as_mts", "=", True)]
-        if self.company_id:
-            wh_domain.append(("company_id", "=", self.company_id.id))
-        wh = wh_obj.search(wh_domain)
-        wh_orderpoint = self.orderpoint_ids.warehouse_id
-        wh_wo_orderpoint = wh - wh_orderpoint
-        if wh_wo_orderpoint:
-            res = wh_wo_orderpoint
-        return res
+    def _update_mto_as_mts_orderpoint(self, original_mto_products):
```
ivantodorovich: I find this method a bit confusing

The fact that's called on a recordset, but in fact acts like an @api.model method is misleading.

IMO the identification of changed products could be factorized to either another method, or just done in the caller.. and then have smaller recordset methods with clear intentions

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -65,40 +51,106 @@ def _prepare_missing_orderpoint_vals(self, warehouse):
...
+        )
+        if not products_by_company:
+            return
+        wh_obj = self.env["stock.warehouse"]
+        op_obj = self.env["stock.warehouse.orderpoint"]
+        all_mto_wh = wh_obj.search([("mto_as_mts", "=", True)])
+        op_vals_list = []
+        for company, products in products_by_company:
+            if company:
+                mto_wh = all_mto_wh.filtered(
```
ivantodorovich: Do a group before entering the loop

`mto_wh_by_company = all_mto_wh.grouped('company_id')`

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -65,40 +51,106 @@ def _prepare_missing_orderpoint_vals(self, warehouse):
...
+        chunck_size = models.INSERT_BATCH_SIZE * 20
+        _logger.debug(
+            f"Create orderpoint for MTO: {len(op_vals_list)} to create - "
+            f"{len(op_vals_list)//chunck_size+1} chucks"
+        )
+        for i, op_vals_list_chunk in enumerate(split_every(chunck_size, op_vals_list)):
+            _logger.debug(f"Create orderpoint for MTO - chunck {i}")
+            ops = op_obj.create(op_vals_list_chunk)
+            # free memory usage
+            ops.invalidate_model()
```
ivantodorovich, line 104: Does it actually improve something?

AFAICS the only thing in the loop is `op_obj.create(op_vals_list_chunk)`, and create is flushed immediately so there's nothing really cached.

You'd probably get better peformance not spliting at all, and letting the ORM created everything on a single query

IMO this pattern only makes sense when you're reading & writing on records:
- Reading (specially with prefetching) can easily consume lots of RAM as you keep iterating on records
- Writing (without flushing) can also do the same, on a much smaller scale

stock_orderpoint_mto_as_mts/models/product_template.py
```diff
@@ -10,31 +10,21 @@ class ProductTemplate(models.Model):
...
         template = super().create(vals_list)
-        template.product_variant_ids.sudo()._ensure_default_orderpoint_for_mto()
+        template.product_variant_ids._ensure_default_orderpoint_for_mto()
         return template
 
     def write(self, vals):
-        original_mto_templates = self.filtered(lambda temp: temp.is_mto)
+        if self.env.context.get("orderpoint_mto_as_mts"):
+            return super().write(vals)
+        original_mto_products = self.product_variant_ids.filtered("is_mto")
```
ivantodorovich: Add a condition to do it only `if "route_ids" in vals or "categ_id" in vals`. which is when we actually need this

stock_orderpoint_mto_as_mts/models/product_category.py
jbaudoux: There can be really a lot of products under that category. Showing some progress in the debug log is convenient.

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -0,0 +1,157 @@
...
+                    warehouse, products.browse()
+                )
+                for product in missing_products:
+                    op_vals_list.append(
+                        product._prepare_missing_orderpoint_vals(warehouse)
+                    )
+        chunck_size = models.INSERT_BATCH_SIZE * 20
+        _logger.debug(
+            f"Create orderpoint for MTO: {len(op_vals_list)} to create - "
+            f"{len(op_vals_list)//chunck_size+1} chucks"
```
jbaudoux: chunk and not chunck
```suggestion
            f"{len(op_vals_list)//chunk_size+1} chunks"
```

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: What I measured is that increasing the chunk size is not making it faster

ivantodorovich: My suggestion is actually to just don't split in chunks, and my point is that splitting and processing in chunks here seems pointless

stock_orderpoint_mto_as_mts/models/product_category.py
ivantodorovich: Hm yes, but there seems to be a lot of them. It's a bit messy, but maybe it's just me.

stock_orderpoint_mto_as_mts/models/stock_route.py
jbaudoux: fixed

stock_orderpoint_mto_as_mts/tests/test_stock_warehouse_mto_as_mts.py
jbaudoux: fixed

stock_orderpoint_mto_as_mts/models/product_category.py
jbaudoux: fixed

stock_orderpoint_mto_as_mts/models/product_category.py
jbaudoux: I compared `filtered_domain` and `filtered` and it's resolved the same way by odoo. It's making the same `SELECT` in db.

I need all the products to ensure the orderpoints. I tried to merge in the previous search that is marking the products as mto but then I'm missing all the records that have their id in the cache as they are recomputed and not returned by the search. To solve that, I should find those records and mark as to not recompute which would be a bit unsafe and unusual. So I keep it as 2 searches

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: filtered domain is making a search limited to the records subset. I have the feeling I'm doing what you suggest as second point?

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: Indeed, I'm dropping that and just keep one method to populate orderpoints and one to archive

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: integrated

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: the benefit it too see it's progressing in the logs :/

stock_orderpoint_mto_as_mts/models/product_template.py
jbaudoux: added

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: `invalidate_model` may still be useful is there are computed fields that are triggered after create, isn't it?

jbaudoux:
@ivantodorovich Thanks a lot for your review. I tried to improve the code. Can you drop another review before I squash the fixup commits ?

stock_orderpoint_mto_as_mts/models/product_product.py
ivantodorovich: Yeah, but, I mean.. KISS
Anyway, not blocking over this

stock_orderpoint_mto_as_mts/models/product_category.py
ivantodorovich: > I compared filtered_domain and filtered and it's resolved the same way by odoo. It's making the same SELECT in db.

Neither `filtered` nor `filtered_domain` filter through SQL.
The select you see is likely the ORM fetching the fields needed to filter through python, which is why it's the same for both.

`filtered_domain`'s implementation is much more complex due to the algorithm needed to handle domains. It shouldn't be a preferred option over `filtered`, unless you're dealing with a domain that's coming from somewhere else and so you need the domain interpretation machinery

In any case, not blocking over this micro optimization

stock_orderpoint_mto_as_mts/models/product_product.py
ivantodorovich: `filtered_domain` is not doing a search, it's fetching all records data and filtering in python. Maybe that's the confusion?

ivantodorovich (approved):

stock_orderpoint_mto_as_mts/models/product_product.py
jbaudoux: oh indeed, I wrongly interpreted that method. I'm replacing by a search !

sebalix (approved):

jbaudoux:
I removed the wrong .pot file

PR #41, opened 2025-06-09
[18.0] stock_orderpoint_default_location: imp test perf

simahawk:
Disable all trackings.

NOTE: not using BaseCommon because it does too many things.


jbaudoux (approved):

rousseldenis:
> Disable all trackings.
> 
> NOTE: not using BaseCommon because it does too many things.

In fact, it does not create anything but initialize some variables:

https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/tests/common.py#L23

rousseldenis (approved):

PR #44, opened 2025-06-13
[18.0][ADD] purchase_stock_product_replenish_supplier: Set default supplier on product replenish

jbaudoux:
Set default supplier on product "Replenish" wizard.

When an orderpoint exists, odoo doesn't set the default supplier in the "Replenish" wizard if not set on the orderpoint. But it is not required to set a supplier on an orderpoint (the field is even hidden by default), it will default to the main product supplier.

Also ensure the supplier is right for the variant.

Relates to https://github.com/odoo/odoo/blob/18.0/addons/purchase_stock/wizard/product_replenish.py#L12

Note: odoo just changed code last week https://github.com/odoo/odoo/commit/6c7d986d4c75419cb8804103b2b97cbc3a9ef361, now the orderpoint should have a route to have a default supplier. But the route is not set by default on an orderpoint. So this module is still relevant.
Related last odoo PR https://github.com/odoo/odoo/pull/210336


rousseldenis (approved):
LGTM

PR #49, opened 2025-06-27
[16.0][IMP] stock_orderpoint_mto_as_mts: Ensure only auto orderpoints getting created and archived

### LINKED ISSUE #47: stock_orderpoint_mto_as_mts based on auto triggered orderpoints
As i checked the new version of `sale_stock_mto_as_mts_orderpoint` because i wanted to add there some new features. I searched for the old method `_run_orderpoints_for_mto_products` i didn't find it.

So is it correct that the addon stock_orderpoint_mto_as_mts is based on the fact that there must be an auto triggered order point? Which will be then be triggered by https://github.com/odoo/odoo/blob/44f9d161234e6f99477f4dc946318c36360ae009/addons/stock/models/stock_move.py#L2132 ?

If that is the case, then I see the following errors that can occur.

- `_trigger_scheduler` can not find a suitable orderpoint
**Either because of**
the default value for the field `trigger` https://github.com/odoo/odoo/blob/44f9d161234e6f99477f4dc946318c36360ae009/addons/stock/models/stock_orderpoint.py#L42 is changed to `manual`  then a `manual` orderpoint is created because this https://github.com/OCA/stock-logistics-orderpoint/blob/183a95fb1eda38e4b91c3225d2628e48b61fb196/stock_orderpoint_mto_as_mts/models/product_product.py#L16 returns no value for `trigger`
**or because** 
a previous created auto triggered orderpoint is changed to `manual` then there is no logic which will ensure that there is a auto triggered orderpoint

@jbaudoux can you please correct me if I am wrong.

mt-software-de:
Closes #47 
cc @jbaudoux

### DIFF
--- a/stock_orderpoint_mto_as_mts/models/product_product.py
+++ b/stock_orderpoint_mto_as_mts/models/product_product.py
@@ -1,4 +1,5 @@
 # Copyright 2023 ACSONE SA/NV
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
@@ -27,6 +28,7 @@ class ProductProduct(models.Model):
         self.ensure_one()
         orderpoints = self.env["stock.warehouse.orderpoint"]
         orderpoint_obj = self.env["stock.warehouse.orderpoint"]
+        orderpoint_vals_base = self._prepare_orderpoint_vals_base()
         for warehouse in warehouses:
             orderpoint = orderpoint_obj.with_context(active_test=False).search(
                 [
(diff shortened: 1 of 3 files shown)


jbaudoux (approved):

jbaudoux (commented):
The line where it unarchive and writes 
`{"active": True, "product_min_qty": 0.0, "product_max_qty": 0.0}`
it should also write `"trigger": "auto"`

stock_orderpoint_mto_as_mts/models/product_product.py
```diff
@@ -27,6 +28,7 @@ def _create_default_orderpoint_for_mto(self, warehouses):
         self.ensure_one()
         orderpoints = self.env["stock.warehouse.orderpoint"]
         orderpoint_obj = self.env["stock.warehouse.orderpoint"]
+        orderponit_vals_base = self._prepare_orderpoint_vals_base()
```
jbaudoux: ponit :)

jbaudoux (approved):

PR #52, opened 2025-06-30
[16.0][IMP] stock_location_orderpoint: returns the ongoing replenishment moves.

nicolas-delbovier-acsone:
Make sure the replenishment method always returns its ongoing replenishment moves.

This is useful in case we want to add custom logic to the generated/updated replenishment moves. If the moves get returned, we do not need to make a second query to gather them afterwards.

This PR is based on #54

### DIFF
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -466,15 +466,17 @@ class StockLocationOrderpoint(models.Model):
     def run_replenishment(self, products=False):
         """Run the replenishment for all potential products or only a selection"""
         procurements = self._prepare_procurements(products)
-        if not procurements:
-            return
-        self.env["procurement.group"].with_context(from_orderpoint=True).run(
-            procurements, raise_user_error=False
-        )
-        self._after_replenishment()
+        if procurements:
+            self.env["procurement.group"].with_context(from_orderpoint=True).run(
+                procurements, raise_user_error=False
+            )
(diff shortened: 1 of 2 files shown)


nicolas-delbovier-acsone:
I forced push to rebase to #54

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -466,33 +466,43 @@ def _prepare_procurements(self, products=False):
...
-            return
-        self.env["procurement.group"].with_context(from_orderpoint=True).run(
-            procurements, raise_user_error=False
-        )
-        self._after_replenishment()
+        if procurements:
+            self.env["procurement.group"].with_context(from_orderpoint=True).run(
+                procurements, raise_user_error=False
+            )
+        return self._after_replenishment()
```
mt-software-de: I would rather move the logic for getting the current replenishment moves into here and then passing them on downwards, instead of getting them from the result of `_after_replenishment`. 

```suggestion
        replenishment_moves = self._get_current_replenishment_moves()
        self._after_replenishment(replenishment_moves)
        return replenishment_moves
```

mt-software-de (changes requested):
Thanks for your contribution, this is a good idea.

stock_location_orderpoint/models/stock_location_orderpoint.py
nicolas-delbovier-acsone: You are right, this is probably better to put this logic there, I'll make a commit

PR #54, opened 2025-07-31
 [FIX] stock_location_orderpoint: fix date dependend test.

nicolas-delbovier-acsone:
Some test was failing because the code saying "day +1" fails if we are the last day of a month.


mt-software-de (approved):
Thx. LGTM

rousseldenis (approved):

PR #20, opened 2024-06-03
[16.0][ADD] stock_location_orderpoint_change_priority

rousseldenis:

PR #55, opened 2025-07-31
[16.0][ADD] stock_location_orderpoint_change_priority V2

nicolas-delbovier-acsone:
This PR is a rework of #20 now based on #52.

Using the fact that replenishment moves are now returned by the replenishment method (#52), this rework allows to simplify the code. 

Compared to #20, the logic changed a bit: Beforehand, the priority update would only target some moves in case of 'auto' orderpoint whereas it would update all ongoing moves in case of 'manual' orderpoint (see the removed unit test). The logic is now simpler: always update all ongoing replenishment moves priority when replenishing.


stock_location_orderpoint/tests/test_location_orderpoint.py
```diff
@@ -96,12 +98,12 @@ def test_cron_replenishment(self):
         self.assertFalse(replenish_move)
 
         # create the available quantity -> replenishment
-        tomorrow = now.replace(day=now.day + 1)
+        tomorrow = now + get_timedelta(1, "day")
```
jbaudoux, line 101:
```suggestion
        tomorrow = fields.Datetime.add(now, days=1)
```

stock_location_orderpoint/tests/test_location_orderpoint.py
```diff
@@ -96,12 +98,12 @@ def test_cron_replenishment(self):
...
         with self._freeze_time(tomorrow):
             self._set_qty_in_location(self.product, location_src, 12)
 
         self.product.invalidate_recordset()
-        day_after_tomorrow = tomorrow.replace(day=tomorrow.day + 1)
+        day_after_tomorrow = now + get_timedelta(2, "day")
```
jbaudoux, line 106:
```suggestion
        day_after_tomorrow = fields.Datetime.add(now, days=2)
```

jbaudoux (commented):
prefer this instead

stock_location_orderpoint_change_priority/readme/CONTEXT.md
```diff
@@ -0,0 +1,8 @@
+Orderpoints have generated replenishments that have a normal priority (e.g.:
+replenishments to cover average sales).
```
jbaudoux, line 2:
```suggestion
replenishments to cover average consumption).
```

stock_location_orderpoint_change_priority/models/stock_location_orderpoint.py
```diff
@@ -0,0 +1,19 @@
...
+class StockLocationOrderpoint(models.Model):
+    _inherit = "stock.location.orderpoint"
+
+    def run_replenishment(self, products=False):
+        replenishment_moves = super().run_replenishment(products)
+        for move in replenishment_moves:
+            # Also update the priority of the picking because the move priority
+            # is a computed field based on that
+            move.picking_id.with_context(
+                no_check_priority=True
```
jbaudoux, line 16: Unknown context variable. Maybe ensure sudo and don't block when called with sudo?

stock_location_orderpoint_change_priority/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Location Orderpoint Change Priority",
```
jbaudoux, line 5: This should be part of the base module. It's a base feature to support mutiple orderpoints with different priorities.

I see @mt-software-de made the same comment :)

jbaudoux (changes requested):

lmignon:
replaced by #82

lmignon:
@nicolas-delbovier-acsone Can you close this one plz?

PR #56, opened 2025-08-19
[18.0][MIG] stock_orderpoint_uom: Migration to 18.0

ArnauCForgeFlow:
Standard Migration to 18.0.

This module comes from stock-logistics-warehouse repo. https://github.com/OCA/stock-logistics-warehouse/tree/15.0/stock_orderpoint_uom


LoisRForgeFlow (changes requested):
Tested in runbot. The procurement UoM is not used when manually procuring:

image

LoisRForgeFlow (approved):
After internal discussion, this is correct, the procurement is done in the procurement UoM but when the result is a PO it is converted to the purchase UoM. In a transfer for instance the proc UoM is kept:

image

AaronHForgeFlow (approved):
Code LGTM

PR #58, opened 2025-11-10
[18.0][ADD] stock_orderpoint_variant_mto_as_mts

henrybackman:
Glue module between `stock_orderpoint_mto_as_mts` and `product_variant_route_mto` to trigger reordering rules for product variants

Depends on:
* https://github.com/OCA/product-attribute/pull/2131


henrybackman:
@jbaudoux Could you check if I've implemented this glue module according to the specifications or why the tests are failing? Adding this module breaks multiple other tests for other modules but I wasn't able to pinpoint a clear reason for this to propose a fix.

stock_orderpoint_variant_mto_as_mts/readme/CONFIGURE.md
```diff
@@ -0,0 +1 @@
+
```
sebalix: Empty file to remove (same for `USAGE.md`)

jbaudoux:
@henrybackman There should be one last test failing. I added this https://github.com/camptocamp/stock-logistics-orderpoint/blob/ea5fb1aadd5144aeac8425074563c691988a9866/stock_orderpoint_variant_mto_as_mts/models/stock_route.py

henrybackman:
> @henrybackman There should be one last test failing. I added this https://github.com/camptocamp/stock-logistics-orderpoint/blob/ea5fb1aadd5144aeac8425074563c691988a9866/stock_orderpoint_variant_mto_as_mts/models/stock_route.py

@jbaudoux There's 4 tests still failing

stock_orderpoint_variant_mto_as_mts/tests/test_product_product.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    def test_orderpoint_with_product(self):
+        # Create orderpoint
+        product = self.env["product.product"].create(
+            {
+                "name": "Test MTO",
+                "type": "consu",
+                "is_storable": True,
+                "route_ids": [(6, 0, [self.mto_route.id])],
```
jbaudoux, line 45:
```suggestion
                "route_ids": [Command.link(self.mto_route.id)],
```

sebalix:
PR to rebase (dependency merged)

jbaudoux:
> PR to rebase (dependency merged)

It's missing post install tag

TDu:
Rebased

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
```diff
@@ -0,0 +1,25 @@
...
+            [("categ_id", "child_of", self.categ_ids.ids)]
+        )
+        products = products.with_context(orderpoint_mto_as_mts=True)
+        if vals["is_mto"]:
+            products.is_mto = True
+        else:
+            mto_routes = self.env["stock.route"].search([("is_mto", "=", True)])
+            if not mto_routes:
+                # When the last mto route is unflagged
+                products.is_mto = False
```
mmequignon, line 24: I'm not sure I understand this right.
We are looking for routes that aren't MTO, but we arent filtering based on product.route_ids.
He're my thinking:

```
mto_routes = self.env["stock.route"].search([("is_mto", "=", True)])
1, 2 # routes with id 1 and 2 are MTO
products.route_ids.is_mto
False # However none of the routes on products are MTO
```
In this case we wouldn't set is_mto False on products, and I believe we should.
Am I missing something here ?
@henrybackman @jbaudoux

mmequignon (changes requested):
If I'm right in my assumption, we are missing something

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
```diff
@@ -0,0 +1,25 @@
...
+        products = self.product_ids.product_variant_ids
+        products |= self.env["product.product"].search(
+            [("categ_id", "child_of", self.categ_ids.ids)]
+        )
+        products = products.with_context(orderpoint_mto_as_mts=True)
+        if vals["is_mto"]:
+            products.is_mto = True
+        else:
+            mto_routes = self.env["stock.route"].search([("is_mto", "=", True)])
+            if not mto_routes:
```
mmequignon, line 22: here, `self.is_mto` is True since `vals["is_mto"]` is False.
mto_routes should always contain self.
This condition will never be met.
`super()` has to be called before, or self excluded from the search

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
```diff
@@ -0,0 +1,25 @@
...
+                # When the last mto route is unflagged
+                products.is_mto = False
+        return super().write(vals)
```
mmequignon, line 25:
```suggestion
        res = super().write(vals)
        if vals["is_mto"]:
            mto_products = products.filtered(lambda p: any(p.route_ids.mapped("is_mto")))
            mto_products.is_mto = True
        else:             
            not_mto_products = products.filtered(lambda p: not any(p.route_ids.mapped("is_mto")))
            not_mto_products.is_mto = False
        return res
```

I was thinking about smth like that

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
TDu: Good point

stock_orderpoint_variant_mto_as_mts/tests/test_product_product.py
henrybackman: Fixed

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
henrybackman: Thanks for catching that! I've reworked this to take that into account and remove the self records from the record set when searching for the remaining mto routes

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
henrybackman: Thank you! This helped a lot :pray: It didn't end up being exactly this but should be working now

stock_orderpoint_variant_mto_as_mts/models/stock_route.py
henrybackman: The `route_ids` should contain all MTO routes so I don't think it is necessary to filter based on product route_ids. Here in the `_compute_route_ids` all MTO routes are added:
- https://github.com/OCA/product-attribute/blob/eaa42bddc45a212f7ab72309c52b64b6d95c3de3/product_variant_route_mto/models/product_product.py#L41

henrybackman:
I added another fix on product-attribute that resolves issue with the is_mto on variant creation that surfaced when resolving the failing tests. This is added to the test-requirements:
- https://github.com/OCA/product-attribute/pull/2397

TDu (approved):
@henrybackman  The dependency has been merged.
@jbaudoux  Could you review again.

PR #66, opened 2026-02-03
[19.0] [ADD] stock_orderpoint_safety_stock

ivantodorovich:
This module enhances inventory management by introducing service-level–based, stochastic control using a continuous review replenishment system. Instead of relying solely on fixed min/max stock levels, the module estimates your typical (mean) and variable (variance) daily demand from historical data and factors in lead time. These calculations generate a statistically sound safety stock, which becomes your reorder threshold (the “min”). The “max” is set so your inventory will cover expected demand during the entire replenishment period.

Now, you can set replenishment rules (orderpoints) based on the **Cycle Service Level (CSL)**, which reflects the probability of meeting all demand during a replenishment cycle without running out of stock. This provides a practical alternative to manually setting minimum and maximum stock thresholds.

With CSL enabled, safety stock is automatically calculated using your historical daily sales data (both average and standard deviation), gathered over a period you define, plus the orderpoint’s lead time. The system then keeps your reordering rule's min and max levels up-to-date.

Prefer manual control? You can always switch back to the “Manual” mode to specify min and max directly.

## Understanding the Theory: Service Levels & Inventory Management

The backbone of this approach is the Cycle Service Level (CSL), a widely used supply chain metric:

- **CSL Definition:** The chance that your inventory will fully cover demand during a restocking cycle.
- **Example:** A CSL of 95% means that only 5% of cycles will risk a stockout.

This system assumes demand is random (not fixed), so it uses statistical methods:

- **Average daily demand** (μ)
- **Standard deviation of daily demand** (σ)

Because demand can fluctuate during the lead time, safety stock acts as a buffer to reduce the risk of running out of stock.

**Safety stock formula:**

`safety_stock = σ_L × z × g`

Where:

- **σ_L:** Standard deviation of demand over the lead time
- **z:** Z-score for your desired CSL (e.g., 1.65 for 95% CSL)
- **g:** Growth factor (optional, lets you add a margin)

**How min and max are derived:**
- **Minimum (min):** `safety stock`
- **Maximum (max):** `min + (average daily demand × lead time in days)`

**Why does it work?**

- Odoo triggers replenishment whenever stock falls below the min (safety stock).
- The max level is set high enough to cover all expected demand until the next restock.
- Thus, your min acts as a buffer, reducing the risk of running out, and max ensures you don’t overorder.

This makes inventory management both more data-driven and easier to maintain.


stock_orderpoint_safety_stock/models/product_product.py
```diff
@@ -0,0 +1,130 @@
...
+        :return: A dictionary with the aggregated values of the series.
+        """
+        serie_by_product = self._get_daily_demand_serie(warehouse, days)
+        return {
+            product: {
+                # Returned to ease overrides in subclasses
+                "_serie": serie,
+                # Values will be written directly to the orderpoint, must be field names
+                "demand_avg_qty": fmean(serie),
+                "demand_std_dev": stdev(serie),
```
vvrossem: I would add a small comment explaining what `fmean` and `stdev` do

stock_orderpoint_safety_stock/models/stock_cycle_service_level.py
```diff
@@ -0,0 +1,111 @@
...
+    return math.sqrt(2.0) * erfinv(2.0 * p - 1.0)
+
+
+class StockCycleServiceLevel(models.Model):
+    _name = "stock.cycle.service.level"
+    _description = "Cycle Service Level"
+    _rec_name = "display_name"
+    _order = "csl desc"
+
+    csl = fields.Float(
```
vvrossem, line 33: It's a matter of personal "taste", but I'd name this:
```suggestion
    cycle_service_level = fields.Float(
```

vvrossem (approved):
Great job :superhero: :goat:

stock_orderpoint_safety_stock/models/stock_cycle_service_level.py
ivantodorovich: I rather keep it like this
"CSL" is actually a well known acronym to refer to the Cycle Service Level

- https://www.quora.com/What-is-CSL-in-a-supply-chain
- https://es.scribd.com/document/207756177/CSL-Based-on-Shortage-Cost

stock_orderpoint_safety_stock/models/product_product.py
ivantodorovich: Added some comments

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,174 @@
...
+        """Compute the Average Daily Demand and its Standard Deviation."""
+        # Clear existing values
+        self.demand_avg_qty = 0.0
+        self.demand_std_dev = 0.0
+        # Process only orderpoints not using an automatic safety stock method
+        self = self.filtered(lambda rec: rec.safety_stock_method != "manual")
+        if not self:
+            return
+        # Group by warehouse and demand history days to batch read groups of stock moves
+        grouped = self.grouped(lambda rec: (rec.warehouse_id, rec.demand_history_days))
```
jbaudoux, line 109: It would have been better to group by location. You can have multiple orderpoints with different routes for a same product inside a warehouse

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,174 @@
...
+        string="Last modification date of Min and Max quantities from Safety Stock",
+        readonly=True,
+    )
+
+    @api.depends(
+        "safety_stock_method",
+        "demand_history_days",
+        "warehouse_id",
+        "product_id",
+        "product_id.stock_move_ids",
```
jbaudoux, line 101: Recomputed on each new planned move?

stock_orderpoint_safety_stock/models/stock_cycle_service_level.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+    _check_csl_range = models.Constraint(
+        "CHECK(csl >= 0 AND csl < 1)",
+        "The cycle service level must be between 0 and 1, typically 0.90...0.99.",
+    )
+    _check_csl_unique = models.Constraint(
+        "UNIQUE(csl)",
+        "There's already a cycle service level with this value.",
+    )
```
yankinmax, line 77: Don't we need a sanitization and validation check for the floating values entered by user to the `csl`?
What do you think about almost equal floating numbers?

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _apply_safety_stock(self):
+        """Apply the safety stock to the orderpoint min and max quantities"""
+        to_apply = self.filtered(lambda rec: rec.safety_stock_method != "manual")
+        to_apply.safety_stock_update_date = fields.Datetime.now()
+        for rec in to_apply:
+            rec.product_min_qty = rec.safety_stock
+            rec.product_max_qty = rec.product_min_qty + (
+                rec.demand_avg_qty * rec.lead_days
+            )
```
jbaudoux: If you do the math, you'll see min should equal max to work.
Take as example a product you sell 10 every day. The safety is 0 as there is no deviation. So your min=0. You will trigger replenishment when you are out of stock but you will have to wait (lead) days before receiving again the product. That's not good.

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,174 @@
...
+                )
+
+    @api.depends("demand_std_dev", "lead_days")
+    def _compute_demand_lt_std_dev(self):
+        for rec in self:
+            rec.demand_lt_std_dev = rec.demand_std_dev * math.sqrt(rec.lead_days)
+
+    @api.depends("safety_stock_method", "demand_lt_std_dev", "z_score", "growth_factor")
+    def _compute_safety_stock(self):
+        self.safety_stock = 0.0
```
yankinmax: Is it expected to set 0 for `safety_stock_method == "manual"` records too?

jbaudoux (commented):
It will only work for products having moves nearly every days (class-A products).

You'll need different formula for class-C products like you can see an example here https://github.com/OCA/stock-logistics-reporting/blob/16.0/stock_average_daily_sale/models/stock_average_daily_sale.py#L100
Example, you sell a product every 10 days for 10 units. deviation = 3, average = 1. You'll need a huge static z-factor on that product that will conterbalance the dynamic aspect of the approach

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,174 @@
...
+            ("safety_stock_method", "!=", "manual"),
+            ("safety_stock_update_date", "<", "today"),
+        ]
+        records = self.search(domain, order="id", limit=batch_size)
+        records._apply_safety_stock()
+        remaining = self.search_count(domain)
+        self.env["ir.cron"]._commit_progress(
+            processed=len(records),
+            remaining=remaining,
+        )
```
yankinmax, line 228: Nice thanks

yankinmax (approved):
Cool and interesting, thanks

stock_orderpoint_safety_stock/models/stock_cycle_service_level.py
ivantodorovich: Good point, I add `digits=(2,2)` to the field definition, that changes the SQL column from `double precision` to `numeric`, and solve it.

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
ivantodorovich: Yes, the field is computed non-stored so we must assign a value

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
ivantodorovich: Even though you can set up different different routes or destination locations, we compute the "demand" per warehouse -- not per location

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
ivantodorovich: It's non-stored. 
This is just there to automatically invalidate the cache upon new moves, but it's not recomputing anything until it's read.

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
yankinmax: Indeed, thanks

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
ivantodorovich: That's a very good point.

A case like this is very unlikely, but it does prove a fault in the implementation.

We will discuss it with @gaetan-vaujour.

There are a few things to consider, though:
- If we set `min_quantity == max_quantity`, the system would trigger replenishments every day, and we want to avoid it. We need a buffer, a difference betwen `min_quantity` and `max_quantity`.
- We do agree that the `min_quantity` must account for the vendor lead days, so perhaps the solution is something closer to what you suggested, like this `min_quantity = safety_stock + (vendor_lead_days * avg_daily_qty)`.
- However, here we make a distintion between the `vendor_lead_days` and the `lead_days`. The current formula considers the `lead_days`, that's a total which is also adding up the `days_to_purchase` variable.
- That variable could be used to set our buffer, meaning that we'd keep the maximum like it is `max_quantity = (lead_days * avg_daily_qty)`
- So, in practice, the buffer would be `days_to_purchase * avg_daily_qty`, which seems reasonable

ivantodorovich:
I switch the PR back in draft for now, until we resolve this subthread: https://github.com/OCA/stock-logistics-orderpoint/pull/66#discussion_r2768777920

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
jbaudoux: In demand driven MRP:
The safety stock is your red zone covering the variation in the demand.
The yellow zone should cover the lead * ADU. The lead is the vendor lead in case of purchase + security.
The min is red + yellow.
The max is min + green zone. There are multiple formulas for the green zone but usually it is calculated as the largest of: 1) % of average daily usage (ADU) over lead time, 2) Minimum Order Quantity (MOQ), or 3) an imposed minimum order cycle.

jbaudoux: By the way, the most important in my opnion is to properly define the yellow and green zone. Red zone is just an extra buffer and neglectable if the others are not properly sized.

ivantodorovich:
I added a fixup commit with the following changes:
- Added digits on `csl`, solves the constraint issue mentioned by @yankinmax 
- Compute safety stock and stats for orderpoints configured as "Manual", too. As these can be useful informative fields.
- Added a hook method `_get_min_max_quantities_from_safety_stock`, to easily customize how to compute min/max from safety stock in custom modules -- allowing for different strategies to be implemented
- Modified min/max computation as follows:
  - `min_qty = self.safety_stock + (self.demand_avg_qty * self.lead_days)`
  - `max_qty = min_qty + (self.demand_avg_qty * self.days_to_order)`
  - This results in a buffer of ~days_to_order. That's what @jbaudoux calls the green zone.
- Added unit tests for the cases @jbaudoux mentions
  - Product sold 10 units every 10 days
  - Product sold 10 units every day, without variance

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
ivantodorovich: https://github.com/OCA/stock-logistics-orderpoint/pull/66#issuecomment-3855504109

ivantodorovich:
UPDATE:

- `safety_stock = std_dev_over_lead_time * z_score * growth` (unchanged)
- `min_qty = safety_stock + (avg_daily_qty * lead_days * growth)` (unchanged)
- `max_qty = min_qty + (avg_daily_qty * cycle_days * growth)` (new!: `cycle_days`)
- `cycle_days`: new configurable parameter to control the desired number of days between replenishments, controls the gap between min and max.
- Modified the replenisment info wizard to display more information to the user, about how these fields are computed, and make it easy to modify the op settings from there
- Better handling of uom rounding: `safety_stock`, `product_min_qty` and `product_max_qty` are rounded to the uom precision.

Diff against the previous fixup: https://github.com/OCA/stock-logistics-orderpoint/compare/dacd0a59be9342e12affd5ee386eae5a5caab446..b97c9dbf8077416a3366e463f70818a830c7eb19

Diff against original version: https://github.com/OCA/stock-logistics-orderpoint/compare/0a3a416a8f9e9f0effdfe86d60c9731d4a085478..b97c9dbf8077416a3366e463f70818a830c7eb19

stock_orderpoint_safety_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @api.onchange(
+        "product_id",
+        "location_id",
+        "safety_stock_method",
+        "cycle_service_level_id",
+        "growth_factor",
+        "cycle_days",
+    )
+    def _onchange_safety_stock_method_apply(self):
```
jbaudoux, line 208: what about a computed to allow to import orderpoints?

you depend on the location but afterwards you compute based on the warehouse and extract location from the warehouse. It would look more logical to me to base everything on that location

jbaudoux: you also missed lead_days in the depends

ivantodorovich: > what about a computed to allow to import orderpoints?
>
> you depend on the location but afterwards you compute based on the warehouse and extract location from the warehouse. It would look more logical to me to base everything on that location

Sorry, I don't understand. Could you explain with an example? 🙏🏻 

> you also missed lead_days in the depends

`lead_days` is computed non-stored -- and this is an onchange
FTR the onchange here is for UX purposes, as the user would expect to "see" the changes immediately. But there is a cron that runs everyday, too.

jbaudoux: ok, I see you rely on manually calling `action_apply_safety_stock` instead

gaetan-vaujour:
Works functionally. The feature is well implemented, it has been tested with hundreds of products & thousands of stock moves and its still quite fast.

ivantodorovich:
@jbaudoux is this ok to merge?

cyrilmanuel (approved):

gurneyalex (approved):

PR #71, opened 2026-02-12
[16.0][FIX] sale_stock_orderpoint_mto_as_mts: run newly created orderpoints

jbaudoux:
As the creation of orderpoints occurs after the move demand is created, we need to run newly created orderpoint that are in auto trigger

cc @lmignon @nicolas-delbovier-acsone @mt-software-de


nicolas-delbovier-acsone (approved):

lmignon:
@jbaudoux https://github.com/OCA/stock-logistics-orderpoint/pull/74 should solve the merge issue.

PR #72, opened 2026-02-12
[18.0][MIG] stock_orderpoint_generator: Migration to 18.0

clementmbr:
- basic migration
- improve how to make stock.move in tests
- adapt tests to new constrains on `product_min_qty < product_max_qty`
- little fix on `auto_same_values` check


stock_orderpoint_generator/views/orderpoint_template_views.xml
```diff
@@ -0,0 +1,178 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_warehouse_orderpoint_template_list" model="ir.ui.view">
```
metaminux:
```suggestion
    <record id="view_warehouse_orderpoint_template_tree" model="ir.ui.view">
```
It's not recommended to replace "tree" by "list" in XML-IDs during migration

See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#tasks-to-do-in-the-migration

> Replace tree view type by list. This means everywhere in Python, JS or XML where there's a mentioning of this type, but keep XML-IDs that contain tree in its name for avoiding subsequent changes in dependent modules.

stock_orderpoint_generator/tests/test_orderpoint_generator.py
```diff
@@ -0,0 +1,292 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import models
+from odoo.exceptions import UserError
+from odoo.tests import TransactionCase
+
+
+class TestOrderpointGenerator(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
metaminux, line 14:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
This is recommended for tests using TransactionCase, see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#tasks-to-do-in-the-migration

metaminux (changes requested):
Everything is functionally right, even the new reconciliation logic against min/max constraint

Only 2 recommendations from OCA migration guidelines (see my comments inline)

There is also a typo (missing "r" in module's name) in the "first" migration commit message, it should be named :
```
[MIG] stock_orderpoint_generator: Basic Migration to 18.0
```
instead of :
```
[MIG] stock_oderpoint_generator: Basic Migration to 18.0
```

So nothing totally blocking, but those changes would be appreciated

Regards,

stock_orderpoint_generator/tests/test_orderpoint_generator.py
sergiocorato: Now you could use BaseCommon for tests

metaminux:
Hello @clementmbr 

Do you need some help on this one ?

clementmbr:
Hi @metaminux ! Thanks for your assertive feedback :pray: 
I think I fixed everything you listed, hope it's ok now!

metaminux (approved):

rousseldenis:
@clementmbr Thanks for this.

Could you put version change in migration commit and make module name appear in last commit? Thanks

image

clementmbr:
Hi @rousseldenis ! Is it ok like this?

metaminux:
> Hi @rousseldenis ! Is it ok like this?

Hello @clementmbr 

I think what @rousseldenis meant when he asked for the version in the [MIG] commit (b4147f8c0cc97d960a2e8012bd3840cd5c6be520) is to have the `__manifest__.py` file updated for the version in this commit.
For the moment, this change is in the "pre-commit auto fixes" one (ada8984fe8133ce9dbf94f77c022dc064f7903b6)

clementmbr:
@metaminux thanks for the explanation! I will change this if needed...

clementmbr:
@metaminux @rousseldenis done! Everything is OK now?

metaminux:
> @metaminux @rousseldenis done! Everything is OK now?

Yes, LGTM 👌🏼

Issue #60, opened 2025-11-27
[16.0 ]stock_orderpoint_origin: 'purchase.order' object has no attribute 'procurement_group_id'


rvalyi:
## Module
stock_orderpoint_origin
## Describe the bug

```
  File "/odoo/src/addons/stock/wizard/stock_scheduler_compute.py", line 42, in procure_calculation
    self._procure_calculation_orderpoint()
  File "/odoo/src/addons/stock/wizard/stock_scheduler_compute.py", line 36, in _procure_calculation_orderpoint
    self.env['procurement.group'].with_context(allowed_company_ids=cids).run_scheduler(
  File "/odoo/external-src/odoo-usability/stock_usability/models/procurement_group.py", line 26, in run_scheduler
    res = super().run_scheduler(
  File "/odoo/src/addons/stock/models/stock_rule.py", line 590, in run_scheduler
    self._run_scheduler_tasks(use_new_cursor=use_new_cursor, company_id=company_id)
  File "/odoo/src/addons/stock/models/stock_rule.py", line 560, in _run_scheduler_tasks
    orderpoints.sudo()._procure_orderpoint_confirm(use_new_cursor=use_new_cursor, company_id=company_id, raise_user_error=False)
  File "/odoo/external-src/odoo-usability/stock_usability/models/stock_warehouse_orderpoint.py", line 41, in _procure_orderpoint_confirm
    res = super()._procure_orderpoint_confirm(
  File "/odoo/src/addons/stock/models/stock_orderpoint.py", line 568, in _procure_orderpoint_confirm
    self.env['procurement.group'].with_context(from_orderpoint=True).run(procurements, raise_user_error=raise_user_error)
  File "/odoo/external-src/stock-logistics-orderpoint/stock_orderpoint_origin/models/procurement_group.py", line 32, in run
    source_groups = [
  File "/odoo/external-src/stock-logistics-orderpoint/stock_orderpoint_origin/models/procurement_group.py", line 33, in <listcomp>
    x.procurement_group_id if x._name != "stock.picking" else x.group_id
AttributeError: 'purchase.order' object has no attribute 'procurement_group_id'
```

## To Reproduce
version 16.0

1. run the scheduler with some procurements leading to purchase orders

**Expected behavior**

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

PR #80, opened 2026-03-30
[16.0][FIX] stock_orderpoint_origin: fix undef procurement_group_id field

rvalyi:
fix AttributeError: 'purchase.order' object has no attribute 'procurement_group_id' #60

### DIFF
--- a/stock_orderpoint_origin/models/procurement_group.py
+++ b/stock_orderpoint_origin/models/procurement_group.py
@@ -26,11 +26,13 @@ class ProcurementGroup(models.Model):
                 if not line["document_in"] and line["document_out"]:
                     source_docs.add(line["document_out"])
             if source_docs:
-                # stock.picking object doesn't have field procurement_group_id,
-                # so we check if the source document is a picking to use the
-                # correct field (group_id)
+                # stock.picking or purchase.order objects don't have a
+                # procurement_group_id field, so we check if the source document has
+                # a procurement_group_id field or we use group_id otherwise.
                 source_groups = [
-                    x.procurement_group_id if x._name != "stock.picking" else x.group_id
+                    x.procurement_group_id
+                    if "procurement_group_id" in x
+                    else x.group_id
                     for x in source_docs
                 ]
                 source_names = ", ".join([x.name for x in source_docs])


lmignon (approved):

renatonlima (approved):

stock_orderpoint_origin/models/procurement_group.py
```diff
@@ -26,11 +26,13 @@ def run(self, procurements, raise_user_error=True):
...
-                # correct field (group_id)
+                # stock.picking or purchase.order objects don't have a
+                # procurement_group_id field, so we check if the source document has
+                # a procurement_group_id field or we use group_id otherwise.
                 source_groups = [
-                    x.procurement_group_id if x._name != "stock.picking" else x.group_id
+                    x.procurement_group_id
+                    if "procurement_group_id" in x
+                    else x.group_id
                     for x in source_docs
```
SirPyTech, line 36: **suggestion**: I'd start a more structured approach to simplify the list comprehension and prepare for other models to be easily added, like:
```suggestion
                # Some models use another field instead of "procurement_group_id"
                exception_model_to_field = {
                    "purchase.order": "group_id",
                    "stock.picking": "group_id",
                }
                source_groups = [
                    x[exception_model_to_field.get(x._name, "procurement_group_id")]
                    for x in source_docs
```
or even extract `exception_model_to_field` in a method to allow override, but it can be done with a refactoring later, and maybe it's overcomplicating :man_shrugging:

SirPyTech (approved):
Thanks for the PR!
I checked and you are right, purchase orders use a different field too (https://github.com/odoo/odoo/blob/6aca71c9522b3be5b221847e8b005f233d3ea872/addons/purchase_stock/models/purchase.py#L30).

I only reviewed the code and suggested a different implementation, but yours is good already 😄

SirPyTech:
I found this while looking for a solution to a similar problem.
It turned out I needed a different solution, that is https://github.com/OCA/stock-logistics-orderpoint/pull/88, could you please have a look? Thanks!

stock_orderpoint_origin/models/procurement_group.py
rousseldenis: @rvalyi I'm concerned too by the generalization as we are not sure at 100% another module will introduce a source document which is not in the known models and would have no group_id nor procurement_group_id field.

PR #82, opened 2026-04-09
[16.0] stock_location_orderpoint:  Correctly update replenishment move priority

lmignon:
When different orderpoints are defined on the same location, it could result in the creation of several
replenishment moves for the same product. When this happens, if the moves are merged together, the
resulting move will be linked to the orderpoint with the highest priority, ensuring that the move 
inherits the correct priority and is processed accordingly.

supersedes #55 by making it a standard feature

PR #85, opened 2026-04-23
[16.0][IMP] stock_location_orderpoint: refactor of the stock location orderpoint module

### LINKED ISSUE #84: [16.0][IMP] stock_location_orderpoint: decouple trigger from replenishment computation
## Context

The `stock_location_orderpoint` module allows defining replenishment rules at the stock location level.

A replenishment rule currently defines:

* a location scope
* a trigger mode (real-time, periodic, manual)
* a replenishment computation method

The current implementation is heavily based on stock moves (`stock.move`) as the entry point to determine which products should be evaluated.

## Problem

The current design introduces several functional and technical limitations.

### 1. Incorrect behavior when no stock moves occur

For periodic rules, only products involved in stock moves since the last execution are evaluated.

This leads to incorrect situations:

* If planned replenishment pickings are deleted
* And no new stock moves occur
* Then the periodic rule does not generate any replenishment
* Even if stock is below the required level

### 2. Incompatibility with advanced replenishment strategies

The current approach prevents implementing strategies where demand evolves independently of stock moves.

Example:

* Seasonality-based replenishment
* Forecast-driven replenishment

In such cases:

* The replenishment need may change without any stock movement
* But the rule is never triggered → missing replenishment

### 3. Tight coupling and maintainability issues

The current implementation couples:

* trigger logic
* product selection
* replenishment computation

This results in:

* complex and heterogeneous code paths
* difficult extensibility
* duplication of logic across trigger modes

### Key idea

Refactor the module to rely on a common abstraction based on the following triplet:

* **product**
* **forecasted quantity at location**
* **replenishment need**

#### 1. Trigger (when to compute)

The trigger becomes responsible for:

* selecting the **products to evaluate**
* computing their **forecasted quantities** at the given location

#### 2. Replenishment strategy (how to compute)

The replenishment method becomes responsible for:

* computing the **replenishment need**
* based on:

  * product
  * forecasted quantity
  * location

=> This fully decouples *when* the rule is executed from *how* the replenishment is computed.

#### Real-time trigger

* Products come from stock moves
* 👍  unchanged behavior

#### Periodic / Manual trigger

* Products are selected based on **recommended quantity.**
* 👎 no longer dependent on stock moves

This ensures:

* all relevant products are evaluated
* even when no stock movement occurred

### Performance considerations

To avoid performance issues when evaluating all products at a location:

* Forecasted quantities are computed using **SQL queries**
* Avoid relying on ORM-based computations for large datasets

### Functional

* Fixes missing replenishment scenarios
* Enables advanced strategies (e.g. seasonality, forecasting)
* Ensures consistent behavior across trigger modes

### Technical

* Clear separation of concerns
* Improved modularity and extensibility
* Simplified and more homogeneous codebase

### Performance

* Maintains performance through SQL-based computations
* Avoids heavy ORM recomputations

## Migration / Compatibility

* This is a **refactoring** with behavioral changes for periodic/manual triggers:

  * rules are no longer dependent on stock moves
* No data model changes expected (to be confirmed if implementation introduces new fields)

## Future improvements enabled

This refactoring enables:

* pluggable replenishment strategies
* forecast-based or seasonality-based computations
* easier integration with demand planning tools

## Notes

* Real-time behavior remains unchanged
* Focus has been put on preserving performance while improving correctness and extensibility

lmignon:
Clean separation of concerns between the computation of the demand and the execution of the procurement, allowing to implement different strategies to compute the demand and to prioritize the replenishment moves.

Defines a new model stock.location.orderpoint.strategy that is meant to be inherited to implement different strategies to compute the demand. The default strategy is implemented in the stock_location_orderpoint_strategy_fill_up model, and implements the "fill up" strategy, which consists in replenishing the stock until the maximum quantity defined on the orderpoint.

We also put in place an explicit orchestration of the replenishment process in the stock.location.orderpoint model, which splits the process in explicit steps allowing to easily hook into the process and modify the behavior of the module. The steps are:
- select the products to replenish
- compute the demand for each product
- compute available quantity in the location for each product
- compute the procurement quantity for each product
- execute the procurement for products with positive procurement quantity
- post-process the created replenishment moves (e.g. to set their priority according to the strategy)
 
Some of these steps default implementation is delegated to the strategy model, allowing to easily change the behavior of the module by implementing a new strategy and/or overriding specific steps of the process.

fixes #84

* [ ] includes #82 
* [ ] #4 has been adapted

### DIFF
--- a/stock_location_orderpoint/models/stock_location.py
+++ b/stock_location_orderpoint/models/stock_location.py
@@ -75,6 +75,19 @@ class StockLocation(models.Model):
         return expression.AND([domain_move_out_loc, domain])
 
     @api.model
+    def _get_replenishing_moves_domain(self, location_id):
+        """Get the domain to apply on stock.move to get the moves replenishing a location."""
+        (
+            _q,
+            domain_move_in_loc,
+            _ol,
+        ) = self._get_stock_domains(location_id)
+        domain = [
+            ("state", "in", ("confirmed", "assigned", "partially_available")),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
+
+    @api.model
     def _get_replenished_moves_domain(self, location_id):
         """Get the domain to apply on stock.move to get the move having repeished the
         location."""
--- a/stock_location_orderpoint/models/stock_location.py
+++ b/stock_location_orderpoint/models/stock_location.py
@@ -88,18 +88,18 @@ class StockLocation(models.Model):
         ]
         return expression.AND([domain_move_in_loc, domain])
 
-    def _clear_caches(self):
+    def _lop_clear_caches(self):
         self._get_cached_stock_domains.clear_cache(self)
 
     @api.model_create_multi
     def create(self, vals_list):
-        self._clear_caches()
+        self._lop_clear_caches()
         return super().create(vals_list)
 
     def write(self, vals):
-        self._clear_caches()
+        self._lop_clear_caches()
         return super().write(vals)
 
     def unlink(self):
-        self._clear_caches()
+        self._lop_clear_caches()
         return super().unlink()
--- a/stock_location_orderpoint/models/stock_location.py
+++ b/stock_location_orderpoint/models/stock_location.py
@@ -55,47 +55,34 @@ class StockLocation(models.Model):
     def _get_cached_stock_domains(self, location_id):
         return self.env["product.product"]._get_domain_locations_new(location_id)
 
-    def _get_stock_domains(self, location_id):
+    def _get_stock_domains(self):
+        self.ensure_one()
         domain = self.env.context.get("excluded_location_domain", [])
         return self.with_context(
             excluded_location_domain_cache_key=make_hashable(domain)
-        )._get_cached_stock_domains(location_id)
+        )._get_cached_stock_domains(self.id)
 
-    @api.model
-    def _get_consuming_moves_domain(self, location_id):
+    def _get_consuming_moves_domain(self):
         """Get the domain to apply on stock.move to get the consuming moves of a location."""
         (
             _q,
             _il,
             domain_move_out_loc,
-        ) = self._get_stock_domains(location_id)
+        ) = self._get_stock_domains()
         domain = [
             ("state", "in", ("confirmed", "partially_available")),
         ]
         return expression.AND([domain_move_out_loc, domain])
 
     @api.model
-    def _get_replenishing_moves_domain(self, location_id):
-        """Get the domain to apply on stock.move to get the moves replenishing a location."""
-        (
-            _q,
-            domain_move_in_loc,
-            _ol,
-        ) = self._get_stock_domains(location_id)
-        domain = [
-            ("state", "in", ("confirmed", "assigned", "partially_available")),
-        ]
-        return expression.AND([domain_move_in_loc, domain])
-
-    @api.model
-    def _get_replenished_moves_domain(self, location_id):
-        """Get the domain to apply on stock.move to get the move having repeished the
+    def _get_replenished_moves_domain(self):
+        """Get the domain to apply on stock.move to get the move having replenished the
         location."""
         (
             _q,
             domain_move_in_loc,
             _ol,
-        ) = self._get_stock_domains(location_id)
+        ) = self._get_stock_domains()
         domain = [
             ("state", "=", "done"),
         ]
--- a/stock_location_orderpoint/models/stock_location.py
+++ b/stock_location_orderpoint/models/stock_location.py
@@ -1,7 +1,16 @@
 # Copyright 2023 ACSONE SA/NV
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models, tools
+from odoo.osv import expression
 from odoo.tools.safe_eval import safe_eval
+
+
+def make_hashable(val):
+    if isinstance(val, list):
+        return tuple(make_hashable(v) for v in val)
+    if isinstance(val, dict):
+        return tuple(sorted((k, make_hashable(v)) for k, v in val.items()))
+    return val
 
 
 class StockLocation(models.Model):
@@ -40,3 +49,57 @@ class StockLocation(models.Model):
             else:
                 action["context"] = str({"default_location_id": self.id})
         return action
+
+    @api.model
+    @tools.ormcache_context("location_id", keys=("excluded_location_domain_cache_key",))
+    def _get_cached_stock_domains(self, location_id):
+        return self.env["product.product"]._get_domain_locations_new(location_id)
+
+    def _get_stock_domains(self, location_id):
+        domain = self.env.context.get("excluded_location_domain", [])
+        return self.with_context(
+            excluded_location_domain_cache_key=make_hashable(domain)
+        )._get_cached_stock_domains(location_id)
+
+    @api.model
+    def _get_consuming_moves_domain(self, location_id):
+        """Get the domain to apply on stock.move to get the consuming moves of a location."""
+        (
+            _q,
+            _il,
+            domain_move_out_loc,
+        ) = self._get_stock_domains(location_id)
+        domain = [
+            ("state", "in", ("confirmed", "partially_available")),
+        ]
+        return expression.AND([domain_move_out_loc, domain])
+
+    @api.model
+    def _get_replenished_moves_domain(self, location_id):
+        """Get the domain to apply on stock.move to get the move having repeished the
+        location."""
+        (
+            _q,
+            domain_move_in_loc,
+            _ol,
+        ) = self._get_stock_domains(location_id)
+        domain = [
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
+
+    def _clear_caches(self):
+        self._get_cached_stock_domains.clear_cache(self)
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        self._clear_caches()
+        return super().create(vals_list)
+
+    def write(self, vals):
+        self._clear_caches()
+        return super().write(vals)
+
+    def unlink(self):
+        self._clear_caches()
+        return super().unlink()
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -466,33 +466,43 @@ class StockLocationOrderpoint(models.Model):
     def run_replenishment(self, products=False):
         """Run the replenishment for all potential products or only a selection"""
         procurements = self._prepare_procurements(products)
-        if not procurements:
-            return
-        self.env["procurement.group"].with_context(from_orderpoint=True).run(
-            procurements, raise_user_error=False
-        )
-        self._after_replenishment()
+        if procurements:
+            self.env["procurement.group"].with_context(from_orderpoint=True).run(
+                procurements, raise_user_error=False
+            )
+        replenishment_moves = self._get_current_replenishment_moves(products)
+        self._after_replenishment(replenishment_moves)
+        return replenishment_moves
 
-    def _prepare_to_assign_replenishment_move_domain(self):
-        """Returns a domain which selects moves created by a replenishment"""
+    def _get_current_replenishment_moves(self, products=False):
+        """
+        Returns ongoing replenishment moves created or updated by the
+        given orderpoints.
+        """
         domain = [
-            ("state", "in", ["confirmed", "partially_available"]),
+            ("state", "in", ["confirmed", "partially_available", "assigned"]),
             ("procure_method", "=", "make_to_stock"),
             ("location_orderpoint_id", "in", self.ids),
         ]
-        return domain
-
-    def _assign_replenishment_moves(self):
-        """Assigns moves created by the orderpoints"""
-        domain = self._prepare_to_assign_replenishment_move_domain()
-        moves_to_assign = self.env["stock.move"].search(
+        if products:
+            domain = expression.AND([domain, [("product_id", "in", products.ids)]])
+        return self.env["stock.move"].search(
             domain, order="priority desc, date asc, id asc"
         )
+
+    def _assign_replenishment_moves(self, moves_to_assign):
+        """Assigns moves created by the orderpoints"""
         for moves_chunk in split_every(100, moves_to_assign.ids):
             self.env["stock.move"].browse(moves_chunk)._action_assign()
+        return moves_to_assign
 
-    def _after_replenishment(self):
-        self._assign_replenishment_moves()
+    def _after_replenishment(self, replenishment_moves):
+        """
+        Assigns generated replenishment moves.
+        """
+        self._assign_replenishment_moves(
+            replenishment_moves.filtered(lambda m: m.state != "assigned")
+        )
 
     def _prepare_orderpoint_domain_location(self, location_ids, location_field=False):
         """
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -401,21 +401,6 @@ class StockLocationOrderpoint(models.Model):
         orderpoints.write({"last_cron_execution": self.env.cr.now()})
         return res
 
-    @api.model
-    def run_auto_replenishment(self, products, locations, location_field="location_id"):
-        """
-        Run the replenishment for all given products.
-        Selects the right orderpoints by locations and location_field.
-
-        :param products: browse record list of product.product
-        :param locations: browse record list of stock.location
-        :param location_field: should be location_id or location_src_id
-        """
-        if not locations or not products:
-            return
-        orderpoints = self._get_orderpoints("auto", locations, location_field)
-        orderpoints.run_replenishment(products)
-
     def run_replenishment(self, products=None, job_logs=None):
         """
         Public entry point to execute the replenishment rule.
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -503,6 +503,12 @@ class StockLocationOrderpoint(models.Model):
         self._assign_replenishment_moves(
             replenishment_moves.filtered(lambda m: m.state != "assigned")
         )
+        for move in replenishment_moves.sudo():
+            new_priority = move.location_orderpoint_id.priority
+            if move.picking_id.priority != new_priority:
+                move.picking_id.priority = new_priority
+            if move.priority != new_priority:
+                move.priority = new_priority
 
     def _prepare_orderpoint_domain_location(self, location_ids, location_field=False):
         """
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -26,10 +26,9 @@ class StockLocationOrderpoint(models.Model):
     name = fields.Char(
         copy=False,
         required=True,
-        readonly=True,
-        default=lambda self: self.env["ir.sequence"].next_by_code(
-            "stock.location.orderpoint"
-        ),
+        readonly=False,
+        default="/",
+        help="Set to '/' and save if you want a default name to be proposed.",
     )
     company_id = fields.Many2one(
         "res.company",
@@ -815,7 +814,7 @@ class StockLocationOrderpoint(models.Model):
         return moves_to_assign
 
     # -------------------------------------------------------------------------
-    # Cache lifecycle
+    # Cache lifecycle and overrides of create/write/unlink
     # -------------------------------------------------------------------------
 
     def _clear_caches(self):
@@ -828,10 +827,27 @@ class StockLocationOrderpoint(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         self._clear_caches()
-        return super().create(vals_list)
+        val_list_updated = []
+        for vals in vals_list:
+            if "name" not in vals or vals["name"] == "/":
+                val_list_updated.append(
+                    dict(
+                        vals,
+                        name=self.env["ir.sequence"].next_by_code(
+                            "stock.location.orderpoint"
+                        ),
+                    )
+                )
+            else:
+                val_list_updated.append(vals)
+        return super().create(val_list_updated)
 
     def write(self, vals):
         self._clear_caches()
+        if "name" in vals and vals["name"] == "/":
+            vals["name"] = self.env["ir.sequence"].next_by_code(
+                "stock.location.orderpoint"
+            )
         return super().write(vals)
 
     def unlink(self):
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -1,15 +1,19 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# Copyright 2026 ACSONE SA/NV (https://www.acsone.eu)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from collections import defaultdict
 from copy import copy
-from datetime import timedelta
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import ValidationError
 from odoo.osv import expression
-from odoo.tools import float_compare, split_every
+from odoo.tools import split_every
 
+from odoo.addons.queue_job.job import identity_exact
 from odoo.addons.stock.models.stock_move import PROCUREMENT_PRIORITIES
+
+from .stock_location_orderpoint_strategy import StockLocationOrderpointStrategy
+from .tools import MovesTouchTracker
 
 
 class StockLocationOrderpoint(models.Model):
@@ -81,15 +85,30 @@ class StockLocationOrderpoint(models.Model):
     location_src_id = fields.Many2one(
         "stock.location", compute="_compute_location_src_id", store=True
     )
+    replenish_limit_to_free_qty = fields.Boolean(
+        string="Limit replenishment to free quantity at source location",
+        help="If enabled, the replenishment quantity will be "
+        "limited to the free quantity at the source location. ",
+        default=True,
+    )
     active = fields.Boolean(default=True)
 
     last_cron_execution = fields.Datetime(
         help="Last time this orderpoint was processed by the cron",
     )
-
     priority = fields.Selection(
         PROCUREMENT_PRIORITIES,
         default="0",
+    )
+    proc_run_async = fields.Boolean(
+        string="Delay procurement execution",
+        help="If enabled, procurements will be run asynchronously in delayed jobs "
+        "when processing this orderpoint. Only applies to 'cron' and 'manual' triggers. "
+        "If disabled, all procurements will be run in the same transaction. ",
+        compute="_compute_proc_run_async",
+        store=True,
+        readonly=False,
+        precompute=True,
     )
 
     _sql_constraints = [
@@ -99,6 +118,10 @@ class StockLocationOrderpoint(models.Model):
             "The combination of Company, Location, Route and Replenish method must be unique",
         )
     ]
+
+    # -------------------------------------------------------------------------
+    # Validation and computed fields
+    # -------------------------------------------------------------------------
 
     @api.constrains("location_id", "route_id")
     def _check_location_id_route_id(self):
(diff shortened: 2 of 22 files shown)


lmignon:
ping @jbaudoux @rousseldenis

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        4. compute delta
+        5. execute replenishment
+        """
+        result = self.env["stock.move"]
+        self.ensure_one()
+        products = self._get_candidate_products(products)
+        if not products and products is not None:
+            # an empty recordset means that we have computed
+            # that no product is concerned by the replenishment.
+            # otherise a None value means that we don't want
```
jbaudoux:
```suggestion
            # that no products are concerned by the replenishment.
            # Otherwise a None value means that we don't want
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +208,157 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+    @api.model
+    @tools.ormcache("orderpoint_id")
+    def _get_consuming_moves_domain(self, orderpoint_id):
+        (
+            _q,
+            _il,
+            domain_move_out_loc,
+        ) = self._get_dest_domains(orderpoint_id)
+        domain = [
+            ("state", "in", ("waiting", "confirmed", "partially_available")),
```
jbaudoux: I would drop `waiting` as it means there is already a move supplying this one.

stock_location_orderpoint/models/stock_location_orderpoint_strategy_fill_up.py
```diff
@@ -0,0 +1,108 @@
...
+        stock_move_obj._apply_ir_rules(query, "read")
+        from_clause, where_clause, params = query.get_sql()
+        sql = f"""
+            SELECT DISTINCT product_id
+            FROM {from_clause}
+            WHERE {where_clause}
+        """
+        self.env.cr.execute(sql, params)
+        product_ids = [row[0] for row in self.env.cr.fetchall()]
+        return orderpoint._product_model.browse(product_ids)
```
jbaudoux: The default strategy is not setting the context and you set the context after the call to this method
```suggestion
        return env[product.product].browse(product_ids)
```

stock_location_orderpoint/models/stock_location_orderpoint_strategy_fill_up.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @api.model
+    def _compute_demand(self, orderpoint, products) -> dict[int, float]:
+        """
+        Compute demand for the given products according to the fill-up strategy.
+        The demand is computed as the quantity required to ensure that the
+        available quantity (virtual + incoming) is not negative, i.e. to fill
```
jbaudoux:
```suggestion
        virtual available quantity is not negative, i.e. to fill
```

stock_location_orderpoint/models/stock_location_orderpoint_strategy_fill_up.py
```diff
@@ -0,0 +1,108 @@
...
+        up the stock up to 0. according to the quantity available at source
+        location and the quantity already incoming.
+
+        :param orderpoint: stock.location.orderpoint record
+        :param products: product.product recordset
+        """
+        demand_data = {}
+        qties_on_dest = products.with_context(
```
jbaudoux: Can you rename `dest` by `location` here and below as it's an orderpoint on a location, not on a dest location.
```suggestion
        qties_on_location = products.with_context(
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+
+        demand_data = self._compute_demand(products)
+        if not demand_data:
+            return result
+
+        # we filter the products to compute only the quantities
+        # for the products for which we have a demand.
+        products = self._product_model.browse(demand_data.keys())
+
+        qty_available_data = self._compute_available_quantities(products)
```
jbaudoux: Shouldn't this be an option of the orderpoint to limit to the free quantity at the source location?
You may want to place the demand even if it is not there to trigger in cascade another replenishment. Also if the procurement is performed by manufacturing, you will never find the product at source location.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+
+        # If the batch size is 0, we run all procurements in the same transaction.
+        result = self._execute_procurements(procurements)
+        return self._after_replenishment(result)
+
+    def _with_excluded_location_context(self, products):
+        """Ensure product computations use orderpoint's excluded location domain."""
+        if products and "excluded_location_domain" not in products.env.context:
+            return products.with_context(
+                excluded_location_domain=self.stock_excluded_location_domain
```
jbaudoux: where is this used?

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +208,157 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        return self._collect_products_by_orderpoint(
+            moves, orderpoints, self._get_consuming_moves_domain
+        )
+
+    @api.model
+    def _get_from_move_supply(self, moves, trigger="auto"):
+        """
+        Return a dict of products by orderpoint to replenish the demand
+        generated by the given moves. (The given moves are supplying the
+        orderpoint location).
```
jbaudoux:
```suggestion
        orderpoint source location).
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +208,157 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+            ("state", "in", ("waiting", "confirmed", "partially_available")),
+        ]
+        return expression.AND([domain_move_out_loc, domain])
 
     @api.model
-    def _filter_moves_triggering_orderpoints(self, moves, trigger="auto"):
-        """Filters moves that trigger orderpoints"""
+    @tools.ormcache("orderpoint_id")
+    def _get_supplying_moves_domain(self, orderpoint_id):
+        self.browse(orderpoint_id)
```
jbaudoux: ?

stock_location_orderpoint/models/stock_move.py
```diff
@@ -94,7 +79,8 @@ def _action_assign(self, *args, **kwargs):
         # location is decreased. So we need to trigger a check for replenishment
         # on this location IOW if an orderpoint exists for this location as
         # target location and the move has the expected characteristics (state, ...)
-        self._prepare_auto_replenishment_for_outgoing_moves()
+        if self:
+            self._prepare_auto_replenishment_for_outgoing_moves()
```
jbaudoux, line 112: I think this should be in a job

stock_location_orderpoint/models/stock_move.py
```diff
@@ -108,5 +94,26 @@ def _action_done(self, *args, **kwargs):
         # on this location IOW if an orderpoint exists for this location
         # as source location and the move has the expected characteristics
         # (state, ...)
-        moves._prepare_auto_replenishment_for_incoming_moves()
+        if moves:
+            moves._prepare_auto_replenishment_for_incoming_moves()
```
jbaudoux, line 127: I think this should be in a job

jbaudoux (commented):
When a supplying move is canceled, the orderpoints configured in auto (relatime) should be run

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+                [move_out_domain, [("product_id", "in", products.ids)]]
+            )
+        move_out_domain = expression.AND(
+            [
+                move_out_domain,
+                [
+                    (
+                        "state",
+                        "in",
+                        ("waiting", "confirmed", "assigned", "partially_available"),
```
jbaudoux: I would drop `waiting` here also as it will be supplied by another move.
```suggestion
                        ("confirmed", "partially_available", "assigned"),
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -86,11 +88,19 @@ class StockLocationOrderpoint(models.Model):
...
         default="0",
     )
+    proc_run_batch_size = fields.Integer(
+        default=0,
+        string="Procurement Run Batch Size",
+        help="Number of procurements to run in one batch when processing this "
+        "orderpoint. 0 means no batching, i.e. all procurements will be run in the "
+        "same transaction. If set to a positive value, procurements will be run in batches "
+        "of the given size in delayed jobs, which can help to avoid long-running transactions "
+        "and reduce the risk of deadlocks.",
```
jbaudoux:
```suggestion
        "and reduce the risk of concurrent updates.",
```

stock_location_orderpoint/models/stock_location_orderpoint_strategy_fill_up.py
```diff
@@ -0,0 +1,108 @@
...
+        self.env.cr.execute(sql, params)
+        dates_by_product = {row[0]: row[1] for row in self.env.cr.fetchall()}
+        for move in replenishment_moves:
+            if move.product_id.id not in dates_by_product:
+                continue
+            move.date = dates_by_product[move.product_id.id]
+        for picking in replenishment_moves.mapped("picking_id"):
+            picking_date = min(picking.mapped("move_ids.date"))
+            if picking_date < picking.scheduled_date:
+                picking.scheduled_date = picking_date
```
mt-software-de, line 146: If the product can not be listed in the dates, we should not check all pickings.
Also mapped is obsolete for relational fields. 
```suggestion
        picking_change_date_ids = set()
        for move in replenishment_moves:
            if move.product_id.id not in dates_by_product:
                continue
            move.date = dates_by_product[move.product_id.id]
            picking_change_date_ids.add(move.picking_id.id)
        pickings = self.env["stock.picking"].browse(picking_change_date_ids)
        for picking in pickings:
            picking_date = min(picking.move_ids.mapped("date"))
            if picking_date < picking.scheduled_date:
                picking.scheduled_date = picking_date
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +208,157 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        It ensures that the context includes the excluded location domain.
+        """
+        self.ensure_one()
+        return self.env["product.product"].with_context(
+            excluded_location_domain=self.stock_excluded_location_domain
+        )
+
+    @property
+    def _strategy_model(self) -> StockLocationOrderpointStrategy:
+        return self.env[f"stock.location.orderpoint.strategy.{self.replenish_method}"]
```
mt-software-de: Could you please write this explicit?
It makes the code much more readable and searchable. 

```suggestion
        self.ensure_one()
        if self.replenish_method == "fill_up":
            return self.env[f"stock.location.orderpoint.strategy.fill_up"]
        raise ValidationError(_("No strategy class available"))
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+    def run_replenishment(self, products=None):
+        """
+        Public entry point to execute the replenishment rule.
+
+        :param products: optional recordset of product.product
+            If provided, limits the computation scope.
+        """
+        result = self.env["stock.move"]
+        for rule in self:
+            result |= rule._run_replenishment(products=products)
```
mt-software-de:
```suggestion
        for orderpoint in self:
            result |= orderpoint._run_replenishment(products=products)
```

stock_location_orderpoint/models/stock_location_orderpoint_strategy.py
```diff
@@ -0,0 +1,55 @@
...
+        By default, we consider that all stockable products are candidates.
+
+        :param orderpoint: stock.location.orderpoint record
+
+         :return: product.product recordset
+
+        This method is only called if the orderpoint is not triggered for a specific product's
+        recordset
+        """
+        return self.env["product.product"].search([("type", "=", "product")])
```
mt-software-de: IMO this is not good it should always be done by the specific strategy.
Also right now this will be never triggered. And i think you would never call this method by a super call. 
For example. For a use case to fill up your stock on a daily basis at the end of the day.
You are using for cron trigger. There would be something implemented to get the last sold products or anything else but never the all. 


```suggestion
        raise NotImplementedErrror(
            _("The _get_candidate_products method should be implemented in the strategy")
        ) 
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +208,157 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        return self.env[f"stock.location.orderpoint.strategy.{self.replenish_method}"]
+
+    # -------------------------------------------------------------------------
+    # Public triggers and orchestration
+    # -------------------------------------------------------------------------
+
+    @api.model
+    def run_cron_replenishment(self, location_ids=None):
+        orderpoints = self._get_orderpoints(trigger="cron", locations=location_ids)
+        res = orderpoints.run_replenishment(products=False)
```
mt-software-de: The strategy should return the right products. Right now this will make no difference if you call without any products. Because it always checks for the products returned by the strategy
```suggestion
        res = orderpoints.run_replenishment()
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
         """
         if not locations or not products:
             return
-        self = self._get_orderpoints("auto", locations, location_field)
-        self.run_replenishment(products)
+        orderpoints = self._get_orderpoints("auto", locations, location_field)
+        orderpoints.run_replenishment(products)
```
mt-software-de: This isn't used anymore by the code and it is useless now with your changes. 
```suggestion
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        """
+        result = self.env["stock.move"]
+        self.ensure_one()
+        products = self._get_candidate_products(products)
+        if not products and products is not None:
+            # an empty recordset means that we have computed
+            # that no product is concerned by the replenishment.
+            # otherise a None value means that we don't want
+            # to limit the scope of the computation
+            return result
```
mt-software-de, line 479: This is either called with products or without.
If it is called without you want to get the products always from the strategy.
Also the code will not really work or is doing nothing with `products == None`. Because you are computing the demand `_compute_demand` will raise an error if it is `None` instead of a product.product browse. 


```suggestion
        products = self._get_candidate_products(products)
        if not products:
            return result
```

mt-software-de (changes requested):
Thx. For taking the time and refactoring. I like it a lot. :-)

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        """Method to execute procurements in a delayed job for this orderpoint
+
+        :param procurement_values: list namedtuple object expected by the run
+        method of procurement.group
+
+        :return: a recordset of created stock.move
+        """
+        result = self._execute_procurements(procurement_values)
+        self._after_replenishment(result)
+        return f"{len(result)} replenishment moves created for {self.display_name}"
```
mt-software-de: This is not a delayed method. It could be called from anywhere. A method should always be called, by what it is doing and not where it is executed from. 

This should be the called `_execute_procurements` and it should call `_execute_run_procurements` and `_after_replenishment`. Also right now `_execute_procurements` and `_after_replenishment` is always called together.

```suggestion
    def _execute_procurements(self, procurement_values):
        """Method to execute procurements this orderpoint

        :param procurement_values: list namedtuple object expected by the run
        method of procurement.group

        :return: a recordset of created stock.move
        """
        result = self._execute_run_procurements(procurement_values)
        self._after_replenishment(result)
        return f"{len(result)} replenishment moves created for {self.display_name}"
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+            qty,
+            product.uom_id,
+            self.location_id,
+            self.name,
+            self.name,
+            self.company_id,
+            proc_vals,
+        )
+
+    def _execute_procurements(self, procurement_values):
```
mt-software-de:
```suggestion
    def _execute_run_procurements(self, procurement_values):
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        if self.proc_run_batch_size > 0:
+            for procurements_chunk in split_every(
+                self.proc_run_batch_size, procurements
+            ):
+                self._enqueue_execute_procurements(procurements_chunk).delay()
+            return result
+
+        # If the batch size is 0, we run all procurements in the same transaction.
+        result = self._execute_procurements(procurements)
+        return self._after_replenishment(result)
```
mt-software-de:
```suggestion
        return self._execute_procurements(procurements)
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +369,362 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        # If proc_run_batch_size is set to a positive value, we run the
+        # procurements in batches in delayed jobs to avoid long-running transactions
+        # and reduce the risk of deadlocks. Otherwise, we run all procurements in
+        # the same transaction.
+        if self.proc_run_batch_size > 0:
+            for procurements_chunk in split_every(
+                self.proc_run_batch_size, procurements
+            ):
+                self._enqueue_execute_procurements(procurements_chunk).delay()
+            return result
```
mt-software-de, line 527: I don't think it is a good approach to a this logik here i see this more as the base method which is doing the job. Further if the proc_run_batch_size is set on a auto triggered orderpoint, then it will always create a useless second job. 

I think this logic should be separated, this would also lead to only one queue job channel. 

My thoughts about trigger and job and batch. 

Auto -> Never needs batch is already called from a job.
Cron -> Batch is useful, but could be the same as coming from auto a job per product and orderpoint. 
Manual -> Do we really want to run without a job. We never know how many products are impacted and how long it would run. 

Wouldn't it be just easier. To create to always a job per product and orderpoint no matter where it was triggered from? It would ease the code, one queue.job.channel not batch_size needed.
The performance could be decreasing for cron and manual because the demand qty is done just for one product and could be done for several at once. But is this really a problem?

```suggestion
```

Here my approach. 

```
def _run_replenishment(self, products):
    # call of _get_candidate_products could be remove
    ...

def run_replenishment(self, products):
    # without checking batch
    ...

def run_replenishments_in_queue(self, products):
    self.ensure_one()
    # Code from stock.move where run is via queue.job
    # move logic from stock.move to here

def run_replenishments_manual(self):
    for orderpoint in self:
        products = orderpoint._get_candidate_products()
        orderpoint.run_replenishments_in_queue(products)

@api.multi
def cron_run_replenishments(self, location_ids=None):
    # renamed from run_cron_replnishment
    orderpoints = self._get_orderpoints(trigger="cron", locations=location_ids)
    orderpoints.run_replenishments_manual()
    orderpoints.write({"last_cron_execution": self.env.cr.now()})
```

mt-software-de (changes requested):

stock_location_orderpoint/models/stock_location_orderpoint.py
jbaudoux: I also think it should be one job per product if you create jobs. While there could be a lot of products, it should be a limited amount of procurements.

You are also creating a regression with this job as you loop over orderpoints. So when you evaluate the second orderpoint, the supply from the first orderpoint won't be created yet. So you will over replenish. You can't do it that way.

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: ???? @jbaudoux the domain is for moves consuming the location to resupply. 'waiting' is therefore required to take into account moves planned from the location to resupply but waiting for availability...

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: @mt-software-de In some case, the products could be computed at the same time as the demand. That's the case with stock_location_orderpoint_average_daily_sale. see https://github.com/OCA/stock-logistics-orderpoint/pull/4/changes#diff-846561b8022b3d23532f639ddc4e780ba8f955272a635a7920a35e5f81795d5bR19

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: For such scenarios, you could extend this add-on to disable `_compute_available_quantities` and implement your own logic to determine the quantity to procure in `_compute_procurement_qty`. However, this would probably depend on more parameters than I can anticipate here. At least the current implementation provides the required hooks to adapt the behaviour to such a need.

stock_location_orderpoint/models/stock_move.py
lmignon: @jbaudoux IMO, that shouldn’t be the case. Everything has been designed so that the logic used to determine whether a procurement process needs to be initiated has minimal impact without requiring additional database access. In this scenario, a job will only be created if it is thought that a replenishment might be necessary; with your proposal, a job would be created for every move...

stock_location_orderpoint/models/stock_move.py
lmignon: same response as for previous comment

stock_location_orderpoint/models/stock_location_orderpoint.py
jbaudoux: @lmignon waiting is "waiting for another move" meaning you already have a supplying move in mto. You should only consider moves in mts that cannot be reserved. That's the only moves requiring a replenishment.

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: I agree that the overlap between delayed jobs in stock.move and delayable logic in the orderpoint should be avoided. Nevertheless, creating one job per product is not an option, since we could end up with thousands of jobs being created unnecessarily for products with no demand.

lmignon: In my opinion, it is always good practice to batch-process anything that can be handled efficiently in batches, and only split the processing at the point where contention or concurrent updates may occur.

lmignon: > You are also creating a regression with this job as you loop over orderpoints. So when you evaluate the second orderpoint, the supply from the first orderpoint won't be created yet. So you will over replenish. You can't do it that way.

This point should be taken into account

stock_location_orderpoint/models/stock_location_orderpoint.py
jbaudoux: You can easily add an option in the location orderpoint similar to https://github.com/OCA/ddmrp/blob/18.0/ddmrp/models/stock_buffer_profile.py#L61
and move
`qty_available_data = self._compute_available_quantities(products)`
inside `_compute_procurement_qty`

```
    def _compute_procurement_qty(self, demand_data):
        """
        Merge demand and available quantities.

        :return: dict {product_id: procurement_qty}
        """
        if not limit_to_free_qty:
            return demand_data

        qty_available_data = self._compute_available_quantities(products)
        return {
            product_id: min(
                demand_data[product_id], qty_available_data.get(product_id, 0)
            )
            for product_id in demand_data
        }
```

stock_location_orderpoint/models/stock_location_orderpoint.py
jbaudoux: @lmignon What about running the orderpoints bottom-up and exclude in the next run the products for which a procurement as been build? That would prevent to generate multiple times a procurement for a same product

stock_location_orderpoint/models/stock_location_orderpoint.py
jbaudoux: > In my opinion, it is always good practice to batch-process anything that can be handled efficiently in batches, and only split the processing at the point where contention or concurrent updates may occur.

But with one job per product, you can put a indentity key to prevent to have 2 jobs for the same product. Don't forget this can be trigerred on action assign / done of another move. It would be safer

lmignon: @mt-software-de @jbaudoux 

On a stock move for AUTO orderpoint we want to prevent
* unnecessary database writes caused by creating delayed jobs for products and moves that do not impact any orderpoint;
* additional database reads caused by recomputing data that is already available in the cache.

To achieve this:
* we rely on cached values to determine whether a stock move impacts an orderpoint;
* we delay the orderpoint pipeline on a per-product basis.

For other orderpoint types (Manual, Cron), the different stages of the pipeline can be more efficiently processed in batches. In some cases, the candidate products may even be identified at the same time as the demand computation (see #4). In such scenarios, contention only occurs when running the procurement itself.

In my opinion, there is no conceptual reason to merge the two points where part of the orderpoint processing is delayed, since they address different constraints and requirements. What could be done instead is to simplify the delay mechanism by removing the ability to configure the delayed batch size and by preventing delayed procurement execution for AUTO procurements.

lmignon: > Wouldn't it be just easier. To create to always a job per product and orderpoint no matter where it was triggered from? It would ease the code, one queue.job.channel not batch_size needed.
The performance could be decreasing for cron and manual because the demand qty is done just for one product and could be done for several at once. But is this really a problem?

I made a simple test on my production DB for the daily usage strategy triggered by a cron... If I create a job per product candidate, it create more than 10.000 jobs.... but in reality, I only generate less than 300 replenishment...  The performance impact is huge. I could lower the number of candidates by computing products for which stock levels have changed since the last run. But in such a case you've also to take into account the products with expiry date.... 

What bothers me about this approach  is that when I try to narrow down the list of candidate products, I'm applying a logic that assumes a new demand could arise due to a specific activity into the stock. In reality, I end up mixing the concepts of candidate / demand / trigger all over again.

jbaudoux: @lmignon Once you computed the demand and know you have 300 products to process, isn't it reasonable to generate a job per product? In this job, you can compute the proper recommended quantity (_compute_procurement_qty) and trace the result, and then build the procurement.

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: done

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: it's used to compute the qty available...

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: @jbaudoux done

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +221,150 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        for the orderpoint location source and its children locations
+        """
+        orderpoint = self.browse(orderpoint_id)
+        return orderpoint._product_model._get_domain_locations_new(
+            [orderpoint.location_src_id.id]
+        )
 
     @api.model
-    def run_auto_replenishment(self, products, locations, location_field=False):
+    @tools.ormcache("orderpoint_id")
```
jbaudoux: maybe overkill as it's used by a single method that is already cached?

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +221,150 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
+
+    @api.model
+    def _collect_products_by_orderpoint(self, moves, orderpoints, get_domain):
+        """Group impacted products by orderpoint using a domain provider."""
+        products_by_orderpoint = defaultdict(self.env["product.product"].browse)
+        for orderpoint in orderpoints:
+            moves_for_orderpoint = moves.filtered_domain(get_domain(orderpoint.id))
```
jbaudoux:
```suggestion
    def _collect_products_by_orderpoint(self, moves, orderpoints, get_moves_domain):
        """Group impacted products by orderpoint using a domain provider."""
        products_by_orderpoint = defaultdict(self.env["product.product"].browse)
        for orderpoint in orderpoints:
            moves_for_orderpoint = moves.filtered_domain(get_moves_domain(orderpoint.id))
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +221,150 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+            if not moves_for_orderpoint:
```
jbaudoux, line 309: Note for later: here above, we could combine with a product domain defined on the orderpoint

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +221,150 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+            _ol,
+        ) = self._get_src_domains(orderpoint_id)
+        domain = [
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
+
+    @api.model
+    def _collect_products_by_orderpoint(self, moves, orderpoints, get_domain):
+        """Group impacted products by orderpoint using a domain provider."""
```
jbaudoux, line 297: As we consider only consuming moves and I guess it will be refined later in the process. Still need to read further :)
 
```suggestion
        """Group potentialy impacted products by orderpoint using a domain move provider."""
```

jbaudoux (commented):
review to be continued...

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,39 +221,150 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        products_by_orderpoint = defaultdict(self.env["product.product"].browse)
+        for orderpoint in orderpoints:
+            moves_for_orderpoint = moves.filtered_domain(get_domain(orderpoint.id))
+            if not moves_for_orderpoint:
+                continue
+            products_by_orderpoint[orderpoint] |= moves_for_orderpoint.mapped(
```
jbaudoux: no need for mapped

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -594,36 +375,548 @@ def run_auto_replenishment(self, products, locations, location_field=False):
...
+        Resolve the product scope.
+
+        Priority:
+        1. explicit products
+        2. strategy override
+        3. default (rule-based #to be implemented if needed)
+
+        :return: product.product recordset
+        """
+        return products or self._strategy_model._get_candidate_products(self)
```
jbaudoux: You should give the products list to the strategy. The strategy should return a new list where the explicit list is reduced by the strategy.

jbaudoux: Note for later: the orderpoint should provide a product domain to limit the scope

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +222,568 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        :return: stock.location.replenishment.computer in-memory record
+        """
+        self.ensure_one()
+        computer = self.env["stock.location.replenishment.computer"].new(
+            {
+                "location_id": self.location_id.id,
+                "location_src_id": self.location_src_id.id,
+                "replenish_limit_to_free_qty": self.replenish_limit_to_free_qty,
+                "excluded_location_domain": self.stock_excluded_location_domain,
+            }
```
jbaudoux: extract in a prepare method as it depends on orderpoint fields that may be extended

lmignon:
@jbaudoux  @mt-software-de 

Into the last commit I added a replenishment_horizon on the orderpoint info. 

When computing a replenishment, it is important to exclude from the outgoing quantities the quantities that will themselves be replenished by a future replenishment. 

When replenishment is performed on a regular basis, only outgoing moves occurring before the next planned replenishment should be taken into account.

jbaudoux (approved):

jbaudoux:
@mt-software-de Can you update your review? It now looks good to me

mt-software-de:
> @mt-software-de Can you update your review? It now looks good to me

Yes i will do.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -22,10 +26,9 @@ class StockLocationOrderpoint(models.Model):
     name = fields.Char(
         copy=False,
         required=True,
-        readonly=True,
-        default=lambda self: self.env["ir.sequence"].next_by_code(
-            "stock.location.orderpoint"
-        ),
+        readonly=False,
+        default="/",
+        help="Set to '/' and save if you want a default name to be proposed.",
```
mt-software-de, line 31: The name should be unique, could you add a sql_constraint?

stock_location_orderpoint/models/stock_location.py
```diff
@@ -40,3 +49,70 @@ def action_open_location_orderpoints(self):
...
+        """Get the domain to apply on stock.move to get the consuming moves of a location."""
+        (
+            _q,
+            _il,
+            domain_move_out_loc,
+        ) = self._get_stock_domains(location_id)
+        domain = [
+            ("state", "in", ("confirmed", "partially_available")),
+        ]
+        return expression.AND([domain_move_out_loc, domain])
```
mt-software-de: Why not use the recordset?

```suggestion
    def _get_stock_domains(self):
        self.ensure_one()
        domain = self.env.context.get("excluded_location_domain", [])
        return self.with_context(
            excluded_location_domain_cache_key=make_hashable(domain)
        )._get_cached_stock_domains(self.id)

    @api.model
    def _get_consuming_moves_domain(self):
        """Get the domain to apply on stock.move to get the consuming moves of a location."""
        (
            _q,
            _il,
            domain_move_out_loc,
        ) = self._get_stock_domains()
        domain = [
            ("state", "in", ("confirmed", "partially_available")),
        ]
        return expression.AND([domain_move_out_loc, domain])
```

stock_location_orderpoint/models/stock_location.py
```diff
@@ -40,3 +49,70 @@ def action_open_location_orderpoints(self):
...
+        location."""
+        (
+            _q,
+            domain_move_in_loc,
+            _ol,
+        ) = self._get_stock_domains(location_id)
+        domain = [
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
```
mt-software-de, line 89: I have not found any of those two methods being used.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+    def _get_consuming_moves_domain(self, orderpoint_id):
+        (
+            _q,
+            _il,
+            domain_move_out_loc,
+        ) = self._get_dest_domains(orderpoint_id)
+        domain = [
+            ("state", "in", ("confirmed", "partially_available")),
+        ]
+        return expression.AND([domain_move_out_loc, domain])
```
mt-software-de: Couldn't you use now the domain from stock.location?
This would ease the code a little bit. 
Since isn't cached i would also switch to use the recordset instead of passing the oderpoint_id. 

```suggestion
    def _get_consuming_moves_domain(self):
        return self.location_id._get_consuming_moves_domain()
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+    def _get_supplying_moves_domain(self, orderpoint_id):
+        (
+            _q,
+            domain_move_in_loc,
+            _ol,
+        ) = self._get_src_domains(orderpoint_id)
+        domain = [
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
```
mt-software-de: Couldn't you use here `_get_replenished_moves_domain` from `stock.location`?

```suggestion
    def _get_supplying_moves_domain(self):
        return self.location_src_id._get_replenished_moves_domain()
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        products_by_orderpoint = defaultdict(self.env["product.product"].browse)
+        for orderpoint in orderpoints:
+            product_domain = orderpoint._get_product_domain()
+            if product_domain:
+                authorized_product = moves.mapped("product_id").filtered_domain(
+                    product_domain
+                )
+                moves = moves.filtered(lambda m: m.product_id in authorized_product)
+            moves_for_orderpoint = moves.filtered_domain(
+                get_moves_domain(orderpoint.id)
```
mt-software-de: To allow to work with an record instead of passing the id. 

```suggestion
                getattr(orderpoint, get_moves_domain)()
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+        :param moves: stock.move recordset
+        :param trigger: the trigger to consider to select the orderpoints to replenish
+        :return: dict {orderpoint: products}
+        """
         # move from location consume order point location -> DO I've to replenish?
         orderpoints = self._get_orderpoints(
             trigger, locations=moves.location_id, location_field="location_id"
         )
+        return self._collect_products_by_orderpoint(
+            moves, orderpoints, self._get_consuming_moves_domain
```
mt-software-de:
```suggestion
            moves, orderpoints, "_get_consuming_moves_domain"
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
-        orderpoints = orderpoints | self._get_orderpoints(
+        orderpoints = self._get_orderpoints(
             trigger, locations=moves.location_dest_id, location_field="location_src_id"
         )
-        if not orderpoints:
-            return self.env["stock.move"]
-        moves = moves.filtered_domain(self._get_moves_domain(orderpoints.ids))
-        return moves
+        return self._collect_products_by_orderpoint(
+            moves, orderpoints, self._get_supplying_moves_domain
```
mt-software-de:
```suggestion
            moves, orderpoints, "_get_supplying_moves_domain"
```

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+    # -------------------------------------------------------------------------
+    # Context-aware model accessors
+    # -------------------------------------------------------------------------
+
+    @property
+    def _product_model(self):
+        """Returns the product model scoped with the orderpoint's excluded
+        location domain."""
+        self.ensure_one()
+        return self.env["product.product"]
```
mt-software-de: Those methods wouldn't be needed anymore when using the methods from stock.location.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -551,79 +251,599 @@ def _get_orderpoints(self, trigger, locations=False, location_field="location_id
...
+
+        # Single call to the computer:
+        #   demand_only=True  → returns {product_id: demand_qty}  (async path)
+        #   demand_only=False → returns {product_id: qty_to_procure} (sync path)
+        computer = self._get_replenishment_computer()
+        procurement_data = computer.compute(
+            products=products,
+            demand_only=is_delayed,
+            job_logs=job_logs,
+        )
```
mt-software-de, line 489: What is the reason to call the compute when it is delayed?
Could you not just call?

`self._enqueue_fulfill_procurement(product_id).delay()`

Otherwise the Computer will compute the demand which isn't needed at this time.

mt-software-de (changes requested):
THX.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -81,16 +84,43 @@ class StockLocationOrderpoint(models.Model):
...
+        help="Additional domain to filter products for this orderpoint. ",
+        default="[]",
+    )
+    replenish_limit_to_free_qty = fields.Boolean(
+        string="Limit replenishment to free quantity at source location",
+        help="If enabled, the replenishment quantity will be "
+        "limited to the free quantity at the source location. ",
+        default=True,
+    )
+    replenish_horizon = fields.Float(
```
jbaudoux, line 98: Since you introduce an horizon, it would be nice to have one for the supply so that replenishment moves planned to far the future are not taken into account (like the horizon on standard orderpoint)

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: this makes the job idempotent

lmignon:
@mt-software-de All changes done

jbaudoux:
@mt-software-de Good for you now? I would like to merge this important work

stock_location_orderpoint/models/stock_location_orderpoint.py
mt-software-de: @lmignon do you want to change this?

stock_location_orderpoint/models/stock_location_orderpoint.py
mt-software-de: Could you please remove this method?

mt-software-de (changes requested):
Just minor changes cleaning up the code. Otherwise than that it looks good to me.

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: @mt-software-de Done

lmignon:
@mt-software-de All changes done

mt-software-de (approved):
Thx.

stock_location_orderpoint/models/stock_location.py
```diff
@@ -40,3 +49,57 @@ def action_open_location_orderpoints(self):
...
+            _q,
+            domain_move_in_loc,
+            _ol,
+        ) = self._get_stock_domains()
+        domain = [
+            ("state", "=", "done"),
+        ]
+        return expression.AND([domain_move_in_loc, domain])
+
+    def _clear_caches(self):
```
rousseldenis: Maybe too generic name in a heavy used model across modules ?

rousseldenis (approved):
LGTM. Minor comment

stock_location_orderpoint/models/stock_location.py
lmignon: @rousseldenis What do you propose?

jbaudoux: prefix with `_lop`

stock_location_orderpoint/README.rst
```diff
@@ -99,6 +99,11 @@ Location Orderpoint configuration
    to avoid confirmed receptions taken into account), fill in the 
    'Domain to filter locations' field.
 
+When different orderpoints are defined on the same location, it could result in the creation of several
```
jbaudoux: @lmignon Can you drop changes to auto generated files.
I think it's in commit 90e193e405a5936fa4d194f8786516bff7708472

lmignon: @jbaudoux done

PR #86, opened 2026-06-02
[16.0][IMP] stock_location_orderpoint: Add manual launch of orderpoints

lmignon:
* [x] includes #85 
In this initial implementation, the lines must be computed / refreshed manually. I could have reset all the lines at the opening of the view but in this case if two people want to work at the same time on the screen it's not possible.... to be discussed.
* [ ] includes #91 

image


lmignon:
ping @jbaudoux ... almost finished...

lmignon:
ping @phschmidt @jbaudoux

PR #90, opened 2026-07-22
[16.0][FIX] stock_location_orderpoint: reclaim moves from lower priority orderpoints 

nicolas-delbovier-acsone:
When running replenishment for a higher-priority orderpoint after a lower-priority
one, existing replenishment moves covering actual demand were
causing the procurement engine to skip updating the necessary moves.

### DIFF
--- a/stock_location_orderpoint/models/stock_location_orderpoint.py
+++ b/stock_location_orderpoint/models/stock_location_orderpoint.py
@@ -432,6 +432,34 @@ class StockLocationOrderpoint(models.Model):
         self.ensure_one()
         return (self.priority, self.location_id.id, product_id)
 
+    def _reclaim_replenishing_moves_from_lower_priority_orderpoints(self):
+        """Reassign existing replenishing moves from lower-priority orderpoints to this one.
+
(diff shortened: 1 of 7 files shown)


stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -454,12 +482,14 @@ def _run_replenishment(self, products=None, processed=None, job_logs=None):
...
-        result = self.env["stock.move"]
+
         processed = processed or set()
         self.ensure_one()
         is_delayed = self._must_delay_fulfill_procurement()
-
         products = self._get_candidate_products(products)
+        replenishment_moves = (
+            self._reclaim_replenishing_moves_from_lower_priority_orderpoints()
+        )
```
lmignon: @nicolas-delbovier-acsone IMO it's semantically wrong to return replenishment moves generated by others orderpoint. This change breaks the api.

lmignon (changes requested):

lmignon:
@nicolas-delbovier-acsone Can you take a look at #97 ? The approach is less intrusive and more direct... (IMO)

PR #91, opened 2026-07-23
[16.0][IMP] stock_location_orderpoint: Cancel fulfilled replenishment moves

lmignon:
when demand becomes zero, and re-run replenishment for the impacted products.


stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -772,6 +803,93 @@ def _assign_replenishment_moves(self, moves_to_assign):
...
+        """Return the domain to search for replenishment moves generated by this
+        orderpoint, for the given products, that have neither started (printed)
+        nor reached a final state."""
+        self.ensure_one()
+        return [
+            ("location_orderpoint_id", "=", self.id),
+            ("product_id", "in", products.ids),
+            ("picking_id.printed", "=", False),
+            ("state", "not in", ("done", "cancel")),
+        ]
```
lmignon, line 905: @jbaudoux To be discussed.... could be 
```
 [
            ("location_orderpoint_id", "=", self.id),
            ("product_id", "in", products.ids),
            ("state", "not in", ("assigned", "done", "cancel")),
        ]
```

lmignon: 🤔

jbaudoux: I would say not in draft, done, waiting, cancel. But I would change to the list of valid states and use IN instead of NOT IN

stock_location_orderpoint/models/stock_move.py
jbaudoux, line 115: As you touch this hook, a side note that action_done is not always called. You could create immediate moves that are created with "done" state. Quite unusal but possible.

divad1196:
@lmignon @jbaudoux That's a great feature, not specific to orderpoints IMO.

Should this be a generic "cancel replenishment" feature that this module could use?
I think it can be useful regardless on how the replenishment was triggered.

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: @jbaudoux `not in draft, done, waiting, cancel`??!! The goal is to retrieve all pickings for which an operator has not yet started processing.

lmignon: @jbaudoux I added "wating" into the list of excluded states

jbaudoux (approved):

jbaudoux:
> @lmignon @jbaudoux That's a great feature, not specific to orderpoints IMO.
> 
> Should this be a generic "cancel replenishment" feature that this module could use? I think it can be useful regardless on how the replenishment was triggered.

@divad1196 It's checking the demand computed by this module, that's why it's inside this module.

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -772,6 +803,109 @@ def _assign_replenishment_moves(self, moves_to_assign):
...
+        self.ensure_one()
+        started_picking_query = self.env["stock.picking"]._search(
+            self._get_not_started_replenishment_pickings_domain(),
+            order="id",
+        )
+        return [
+            ("location_orderpoint_id", "=", self.id),
+            ("product_id", "in", products.ids),
+            ("picking_id", "in", started_picking_query),
+            ("state", "not in", ("waiging", "done", "cancel")),
```
divad1196: critical:
typo: _wai_**g**_ing_

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -772,6 +803,109 @@ def _assign_replenishment_moves(self, moves_to_assign):
...
+        return self.env["stock.move"].search(
+            self._get_not_started_replenishment_moves_domain(products)
+        )
+
+    def _get_not_started_replenishment_moves_domain(self, products):
+        """Return the domain to search for replenishment moves generated by this
+        orderpoint, for the given products, that have neither started (printed)
+        nor reached a final state."""
+        self.ensure_one()
+        started_picking_query = self.env["stock.picking"]._search(
```
divad1196: variable is `started_picking_query` but the method is "**not**_started".
They should match

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -772,6 +803,109 @@ def _assign_replenishment_moves(self, moves_to_assign):
...
+            self._get_not_started_replenishment_pickings_domain(),
+            order="id",
+        )
+        return [
+            ("location_orderpoint_id", "=", self.id),
+            ("product_id", "in", products.ids),
+            ("picking_id", "in", started_picking_query),
```
divad1196: `("picking_id", "in", started_picking_query)`

We are not handling MO here ?

stock_location_orderpoint/models/stock_location_orderpoint.py
```diff
@@ -772,6 +803,109 @@ def _assign_replenishment_moves(self, moves_to_assign):
...
+
+        :param products: product.product recordset to check
+        :return: recordset of cancelled stock.move
+        """
+        self.ensure_one()
+        moves = self._get_not_started_replenishment_moves(products)
+        if not moves:
+            return moves
+        demand_data = self._get_remaining_demand(moves)
+        moves_to_cancel = moves.filtered(lambda m: not demand_data.get(m.product_id.id))
```
divad1196, line 835: So
1. We get all the demand
2. We find moves whose product don't have demand at all
3. We cancel these moves

We don't handle cases where the move's quantity is bigger than the demand ?
I think this can be quite common scenario.
It's fine because it's hard to handle, but worth a TODO comment IMO.

Edit: example, we have a move for 10 units, we cancel a demande of 4 units -> the move's quantity should be reduced. We are not handling this case

divad1196 (changes requested):
One critical typo to fix and 3 other things that should be handle/clarified

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: oh.... good catch @divad1196

stock_location_orderpoint/models/stock_location_orderpoint.py
lmignon: ?MO?

lmignon:
@divad1196 Comments taken into account...

stock_location_orderpoint/models/stock_location_orderpoint.py
divad1196: @lmignon Manufacturing Order moves, they don't have a `picking_id` but can be a source of replenishment.

lmignon: @divad1196  I added an extension point in the latest commit to address your comment. I hope this covers the use case you described.

divad1196: It does, this is great, thank you !

divad1196 (approved):

